PR #82, opened 2014-03-24
Add FeatureRuntimeAttributes

fchaillou:
FeatureRuntimeAttributes are attributes transmitted to ActivationStrategy that are interested in it by implementing ActivationStrategyWithRuntimeAttributes instead of the base ActivationStrategy interface.
It allows the strategy to decide whether the feature should be enabled or not depending on attributes that are not user specific and can change at runtime.

In my company, we are starting to use togglz on backend application with no user and we need to be able to activate/deactive features depending on runtime attributes.
To be more precise, we are processing requests in Camel and we will want to have the feature active depending on data which is part of the request.

For that, we are proposing to introduce FeatureRuntimeAttributes which are attributes transmitted to the ActivationStrategy directly from the feature.

In our case, we envision to pass directly the data from the place where we want to check for the feature activation.

I might need to add some tests to show more in detail how we think about it and see with you if this could be mergeable or the changes you would like.

Thank you
Fabien


fchaillou:
This PR is probably too old to be still useful, i will close it.

PR #122, opened 2015-03-16
Fix for issue 121 so that metadata is created when enums are added to

RichSteele:
EnumBasedFeatureProvider instead of for each call to getMetaData().


chkal:
Hey @RichSteele. Thanks so much. This patch looks fine. I think this may really fix the performance issues you had. Thanks! ;)

PR #123, opened 2015-03-17
Fix for erratic test failure

RichSteele:
Bump up the sleep time to (apparently) resolve the erratic failures.


chkal:
Thanks! :)

Issue #124, opened 2015-04-09
Refactor FileBasedStateRepository to allow for other kinds of non-file based property providers


RichSteele:
See https://groups.google.com/forum/?fromgroups#!topic/togglz-dev/pGk4qGX2bN4

In short, projects wanting to use properties from other sources than files cannot use the FileBasedStateRepository nor extend it, resulting in duplicated logic for parsing the property keys.  We need to extract the property handling portions of FileBasedStateRepository into a new PropertyBasedStateRepository that uses a PropertySource interface (which ReloadablePropertiesFile implements) for getting and setting the property values.

chkal:
PR applied. ;)

PR #125, opened 2015-04-09
Refactored FileBasedStateRepository and ReloadablePropertiesFile to use a new PropertySource to allow for other property-based repositories other than files

RichSteele:
See issue https://github.com/togglz/togglz/issues/124.


chkal:
Looks perfect. Thanks a lot.

PR #132, opened 2015-05-11
Implementation of TogglzRuie with variations (as per #131)

### LINKED ISSUE #131: [feature idea]: Rework @RunWith(FeatureVariations.class) into a @Rule
I really like the testing support you've implemented in togglz, and the FeatureVariations concept is smart and elegant.  It'd be nice if this could be reworked into a @Rule, though, perhaps an enhancement to TogglzRule?

My main thought was that the TogglzRule#apply(...) could basically call base#evaluate() multiple times, one for each variation.  The logic that works out the different variations (that I presume is part of the FeatureVariations runner) would just move to drive a loop inside of #apply().

The API would be something like:

@Rule
  public TogglzRule togglzRule = TogglzRule.allEnabled(MyFeatures.class)
                                                                        .withVariations()
                                                                         .enable(MyFeatures.F1)
                                                                         .vary(MyFeatures.F2)
                                                                         .vary(MyFeatures.F3);

or something like that.

danhaywood:
turned out to be quite straightforward ... your code was nice n easy to work with!  I believe the tests show it's working ok.

### DIFF
--- a/junit/src/main/java/org/togglz/junit/TogglzRule.java
+++ b/junit/src/main/java/org/togglz/junit/TogglzRule.java
@@ -1,5 +1,7 @@
 package org.togglz.junit;
 
+import java.util.Set;
+
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
@@ -7,6 +9,7 @@ import org.togglz.core.Feature;
 import org.togglz.core.context.FeatureContext;
 import org.togglz.core.util.NamedFeature;
 import org.togglz.core.util.Validate;
+import org.togglz.junit.vary.VariationSetBuilder;
 import org.togglz.testing.TestFeatureManager;
 import org.togglz.testing.TestFeatureManagerProvider;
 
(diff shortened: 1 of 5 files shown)


chkal:
Awesome. The API really looks nice.

I'm wondering if we could get rid of the large if/else in the `apply` method? Could we just remove `initiallyActive` and always use the builder somehow? Wouldn't this simplify the whole code a lot? Just thinking out loud. ;)

danhaywood:
Refactor away... I just did it this way so I could see that every case was
covered.  But if you see a more elegant implementation, go for it.  We have
some tests to verify it works ok :-)

Also, I think there's some redundancy in those tests, probably only my port
of your original variations test [1] and the new CalledMultipleTimes one
[2] are needed; those two "sad cases" ones are probably unnecessary.

Cheers
Dan


[1]
https://github.com/togglz/togglz/blob/master/junit/src/test/java/org/togglz/junit/vary/TogglzRuleWithVariations_HappyCase_Test.java
[2]
https://github.com/togglz/togglz/blob/master/junit/src/test/java/org/togglz/junit/vary/TogglzRuleWithVariations_CalledMultipleTimes_Test.java

chkal:
Hey. I just refactored the TogglzRule a bit. Basically the rule is now always configured using the builder. This means that a standard rule configuration which executes the underlying test only once is now just one "special case". It would be great if you could have a look at the code. I would love to hear your opinion.

danhaywood:
Looks good, sensible rationalization.  Might be worthwhile adding a test to demonstrate/explain what happens when there are variations and also the @WithFeature{Enabled/Disabled} annotation.

chkal:
BTW: In your test there is a comment saying that the order of execution differs between IDE and command line. It would be interesting to see if this can be fixed by changing those two sets to LinkedHashSet:

https://github.com/togglz/togglz/blob/8f42d43f2546ef84314e2087bbf6858e65b68955/junit/src/main/java/org/togglz/junit/vary/VariationSetBuilder.java#L100-101

danhaywood:
That might fix it, though it could also be an issue with reflection.  I
don't know whether Java guarantees that getEnumConstants() always returns
elements in the same order, but if it doesn't then that might account for
the differences.

https://github.com/togglz/togglz/blob/8f42d43f2546ef84314e2087bbf6858e65b68955/junit/src/main/java/org/togglz/junit/vary/VariationSetBuilder.java#L103

chkal:
I agree that there should more tests for this. I'll also try to fix this ordering issue. I also ran into this some time ago. Would be great if the variation order would be more deterministic.

chkal:
I fixed the ordering of the feature variations. The tests are much cleaner now.

PR #133, opened 2015-05-14
Adding test for CassandraStateRepository

artur-tamazian:


chkal:
Awesome. Thank you so much. As I'm not very familiar with Cassandra, having some tests in place that verify the correct behavior feels much better. :)

PR #135, opened 2015-07-08
Fixed activation strategy removing in CassandraStateRepository.

atfire:
There was a problem when you save a feature state with an activation strategy and then remove an activation strategy and save it again. Activation strategy related columns were not deleted from Cassandra.


chkal:
Awesome. Thanks a lot.

PR #143, opened 2015-11-04
Return empty list instead of null to make TestFeatureManager usable w…

BastianVoigt:
…ith Console Servlet


chkal:
Thanks a lot! :)

PR #144, opened 2016-01-06
Add clear ability to cache

sunyal:
Exposing the clear ability allows us to invalidate the cache from a notification


chkal:
Looks good. Thanks a lot.

Issue #151, opened 2016-01-18
Provide more control over the Admin Console's authentication mechanism


marceloverdijk:
Currently the authentication of the Admin Console is tight to the `FeatureManager`s `UserProvider`.

I quick (short-term?) solution would be to make if configurable to completely bypass the the check in the `TogglzConsoleServlet` to see if user is admin (and let apps take care of that in their own way).
This would also be extremely useful when providing a `NoOpUserProvider`. In that case you don't have to provide a custom `UserProvider` to just provide a simple user with admin functionality.

Another (long-term?) solution would be as @chkal suggested is to have a plugable authentication strategy.

Personally I would be fine with option 1 :-)

If interested I can provide a PR which introduces a `is_feature_admin` servlet init-param which can be set to true to by bypass current check from current user.

Another approach would be to make something like `getUserProvider` available in the `FeatureManager` interface, and in case it is a `NoOpUserProvider` the `TogglzConsoleServler` reckons it's alright. Sounds also like a decent solution imo.

Adding this make my Spring Boot Auto Configuration a little bit easier.
When the user does not provide a `UserProvider` I want to create a `NoOpUserProvider` but this causes the admin console unavailable.

chkal:
I just committed a minor change which is basically a very lightweight way for option 2.

Basically users can now simply extend `TogglzConsoleServlet` and overwrite the `isFeatureAdmin` method. I actually think this is the most flexibel solution. IMO using a custom authentication strategy for the console servlet is actually a very rare case. But now users can simply extend the servlet and implement their own strategy. Example:

```java
public class NoAuthTogglzConsoleServlet extends TogglzConsoleServlet {

    @Override 
    protected boolean isFeatureAdmin(HttpServletRequest request) {
        return true;
    }
    
}
```

What do you think? Would this fit your needs?

However, we could think about a more flexibel way of handling the authentication for the admin console which doesn't necessarily require a proper UserProvider.

marceloverdijk:
Yes it's a good start :-)

For me personally a more flexible handling of authentication should not be the scope of Togglz.
I think that's something for the security framework used in the app itself.
Therefore the ability to bypass the `UserProvider` should be enough imo.

marceloverdijk:
I will use something like this for the Boot auto configuration

```
        @Bean
        public ServletRegistrationBean togglzConsole() {
            String path = properties.getConsole().getPath();
            String urlMapping = (path.endsWith("/") ? path + "*" : path + "/*");
[...]

PR #152, opened 2016-01-22
Added easy way of disabling security of admin console

marceloverdijk:
Added easy way of disabling security of admin console.
This is quite useful if security of the console is done via other security system used by application.

### DIFF
--- a/console/src/main/java/org/togglz/console/TogglzConsoleServlet.java
+++ b/console/src/main/java/org/togglz/console/TogglzConsoleServlet.java
@@ -27,12 +27,19 @@ public class TogglzConsoleServlet extends HttpServlet {
 
     protected FeatureManager featureManager;
 
+    protected boolean secured = true;
+
     @Override
     public void init(ServletConfig config) throws ServletException {
 
         featureManager = new LazyResolvingFeatureManager();
 
         servletContext = config.getServletContext();
+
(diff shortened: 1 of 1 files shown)


marceloverdijk:
See https://github.com/togglz/togglz/issues/151

chkal:
Looks good. Thanks a lot.

marceloverdijk:
Thx!

marceloverdijk:
Btw when do you think you will release 2.3.0? With these and the Spring core changes this would be useful.

chkal:
Not sure when 2.3.0 will be ready. I think that it would be good if more people would test your Spring Boot integration. So if anything comes up that would require more changes in Togglz itself, we could get them in quickly.

Not sure what your plans for the Boot Integration module are. But in theory it could become a Togglz module. Or do you want to keep it standalone?

marceloverdijk:
@chkal No I don't want to keep it standalone. However I would prefer it to be merged in directly to Spring Boot itself. I'm in talks with the Boot devs about that.
If they don't want to include it in Boot I think best option is to make it a Togglz module.

chkal:
Ok, sure. I just wanted to emphasize that I would of cause be happy to see it as a Togglz module. However, integrating it into Spring Boot directly may be even better. :smile:

Issue #134, opened 2015-05-19
Add Spring-Boot integration


php-coder:
I'd like to have out-of-box integration with Spring Boot.

chkal:
Could you tell us what you are missing. Basically I assumed that using the `togglz-spring` module should be all you need to get Togglz up and running in Spring Boot applications.

php-coder:
`tooglz-spring` is only enough for "old" Spring applications. When we are using Spring Boot there is a additional ways to reduce configuration even more and make it work out of box (with minimal or zero configuration).

Hm... probably, instead of writing my ideas I will create trivial module that will implement my requirements?

chkal:
Interesting approach. I have a few question though:

* So you have to create a ServletRegistrationBean to automatically register a servlet with Spring Boot? So isn't web-fragment.xml registration not supported at all?

* I think all this Spring Boot specific initialization code could be handled by a SpringBootFeatureManagerProvider. So the initialization would be performed lazily as soon as Togglz is first used. Do you think this could work? Or has such initialization work to be one during the Spring Boot initialization procedure?

csavory:
@chkal usually there is no web.xml or web-fragment.xml in a Spring Boot application.  Most of them will use an Embedded Tomcat (by default) or Jetty if Jetty is on the classpath.

michaelom:
+1

chkal:
Any contributions for improved Spring Boot support are more than welcome. I'm not very familiar with Spring Boot so any help there would be great.

mariuszs:
:+1:

binkley:
I have two bits of sample code I'm working on:
1. Wire vs ignore beans annotated with @ConditionalOnFeature(<enum value>) - only create bean if toggle is enabled
2. Same for individual request handler methods of a Spring controller (see http://binkley.blogspot.com/2016/01/spring-techniques-feature-toggles-for.html for approach but not details)
Would either of these be helpful?  I'm happy to contribute the code, and it's all public domain.

marceloverdijk:
I'm quite interested in this myself and written custom Boot autocofigurations; I will have a look at this.

marceloverdijk:
@chkal would you be interested in a contributed togglz-boot module which acts as a starter and autoconfiguration for Spring boot.
Meaning it will autoconfigure Togglz automatically with sensible defaults but off course this can be changed/overridden by the users.

chkal:
@marceloverdijk & @binkley:

[...]

PR #154, opened 2016-02-09
#134 Add Spring-Boot integration

marceloverdijk:
Here is the Spring Boot integration code.

We need to give some special attention to `spring-boot-starter/template.mf`:
```
..
Import-Template:
 com.github.heneke.thymeleaf.togglz.*;version="0",
 javax.servlet.*;version="0",
 javax.validation.*;version="0",
 org.springframework.*;version="[3.0,4.0)",
 org.springframework.boot.*;version="0",
 org.springframework.security.*;version="[2.0,4.0)",
 org.togglz.*;version="${osgi.bundles.version:default}"
```

I'm not familiar with this bundlor template.mf format.
I took examples from `spring-core` and `spring-security` regarding versions.
For the rest I took `"0"` as I noticed in other files as well.
Does `4.0` mean that it should be in `4.0.x` range? Spring is already in `4.2.x` range so I then wonder if template.mf contents are correct?

I will create another PR in the `togglz-site` repo containing the documentation later.

### DIFF
--- a/spring-boot-starter/pom.xml
+++ b/spring-boot-starter/pom.xml
@@ -1,0 +1,114 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.togglz</groupId>
+    <artifactId>togglz-project</artifactId>
+    <version>2.3.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>togglz-spring-boot-starter</artifactId>
+  <name>Togglz - Spring Boot Starter</name>
+  <description>Togglz - Spring Boot Starter</description>
+
+  <properties>
+    <hamcrest.version>1.3</hamcrest.version>
+    <hibernate-validator.version>5.2.2.Final</hibernate-validator.version>
+    <spring-boot.version>1.3.2.RELEASE</spring-boot.version>
+    <spring-security.version>4.0.3.RELEASE</spring-security.version>
+    <thymeleaf-togglz.version>1.0.1.RELEASE</thymeleaf-togglz.version>
+  </properties>
+
+  <developers>
+    <developer>
+      <id>marceloverdijk</id>
+      <email>maceloverdijk@gmail.com</email>
+      <name>Marcel Overdijk</name>
+      <timezone>CET</timezone>
+    </developer>
+  </developers>
+
+  <dependencies>
+
+    <dependency>
+      <groupId>org.togglz</groupId>
+      <artifactId>togglz-core</artifactId>
+    </dependency>
+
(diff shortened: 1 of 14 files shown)


spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+      <artifactId>hibernate-validator</artifactId>
+      <version>${hibernate-validator.version}</version>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.spec.javax.servlet</groupId>
+      <artifactId>jboss-servlet-api_3.0_spec</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
php-coder: Do we really need this dependency? Is it will be added to any application that uses this module? Hm... Looks like overkill and dependency to something from `org.jboss` maybe isn't good IMHO.

Do we need it for `javax.servlet.*` classes? What dependency Spring Boot uses for that?

spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+
+  <artifactId>togglz-spring-boot-starter</artifactId>
+  <name>Togglz - Spring Boot Starter</name>
+  <description>Togglz - Spring Boot Starter</description>
+
+  <properties>
+    <hamcrest.version>1.3</hamcrest.version>
+    <hibernate-validator.version>5.2.2.Final</hibernate-validator.version>
+    <spring-boot.version>1.3.2.RELEASE</spring-boot.version>
+    <spring-security.version>4.0.3.RELEASE</spring-security.version>
```
php-coder: Spring Boot parent module already has versions for ham crest, hibernate validator and spring security. Could we use dependency from parent module? If not then we will always update them with each new Spring Boot release. Also it's possible to mix different versions of the same dependency... Hm.. Please, correct me if I'm wrong.

marceloverdijk: The parent module is not boot, but Togglz. Also note dependencies are optional so I don't see issue here.
In theory the actual dependnecy will be provided by Boot itself. I just need them to compile the code.

spring-boot-starter/pom.xml
marceloverdijk: It is provided scope so I don't see the issue. This is how Togglz project depends on servlet spec.
It just assumes `javax.servlet.*` is available. Only needed to compile project, and actual servlet implementation provided by Boot (or it's appropriate starter).

chkal: I think `<optional>` dependencies are fine. But I'm curious. Why is there an optional dependency on Hibernate Validator?

Togglz uses the JBoss provided Java EE spec Maven artifacts. Actually it doesn't matter so much which one is used as it is `provided` scoped.

spring-boot-starter/pom.xml
chkal: I agree that the actually version isn't so important because of the optional nature of the dependencies. Basically it is more some kind of "minimal" version that Togglz supports. That's why the Togglz Spring module for example depends on a very old Spring version.

chkal:
@marceloverdijk Thank you very much for the pull request. I think it is fine.

I'll merge it now. Perhaps we get some more people to test as it is now part of the official snapshots.

Thanks a lot for your help. :beers:

PR #155, opened 2016-02-15
Fix travis notifications

kamilszymanski:


chkal:
Good catch. Thanks. :beers:

PR #158, opened 2016-05-02
Use ConcurrentReferenceHashMap instead of blocking 'synchronized'

dstepanov:
During profiling application I have found many threads are at some point waiting for FeatureContext call.
My solution is to use ConcurrentReferenceHashMap class created by Doug Lea and released under public domain.

Many projects are doing the same:
https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/util/ConcurrentReferenceHashMap.java
https://github.com/javagems/grizzly-http-webserver/blob/a47796b1cd6f17919aa405a9bfae7e223a46cef4/src/com/sun/grizzly/util/ConcurrentReferenceHashMap.java

### DIFF
--- a/core/src/main/java/org/togglz/core/context/FeatureContext.java
+++ b/core/src/main/java/org/togglz/core/context/FeatureContext.java
@@ -5,13 +5,16 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ServiceLoader;
-import java.util.WeakHashMap;
 
 import org.togglz.core.logging.Log;
 import org.togglz.core.logging.LogFactory;
 import org.togglz.core.manager.FeatureManager;
 import org.togglz.core.spi.FeatureManagerProvider;
+import org.togglz.core.util.ConcurrentReferenceHashMap;
 import org.togglz.core.util.Weighted;
+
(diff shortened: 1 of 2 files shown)


chkal:
Hey. Thanks a lot for the pull request.

I'm open for any improvement over the current `synchronized` implementation. To be honest, I'm not an expert for tuning such multi-threaded scenarios. So any help here is welcome.

A few things: 
* Your pull request contains many white spaces changes which makes reading it very hard. Could you perhaps fix that? There is a code format profile in the source code repository that you could use. Or simply don't apply your code format after applying the change.
* Do you have any link to the original class by Doug Lea?
* You could move the `ConcurrentReferenceHashMap` to the util package there are some other places where this could be useful.

Awesome! Thanks a lot!

dstepanov:
Hey,

I have added back removed spaces by Idea :smile: added link to the original file and moved `ConcurrentReferenceHashMap` to `util`, it was in the same package to have package scope.

chkal:
Awesome! Thanks a lot. I just merged your patch. :beers:

dstepanov:
Great. Thanks! :smiley: :beers:

PR #159, opened 2016-05-13
Add micro-benchmarks to JMH

ryangardner:
There is a README.md file that describes how to execute these benchmarks. 

I don't have then running automatically as part of the build because they take some time to run - a few minutes each one.

You can either package them up and run them all at once or you can run them individually in an IDE. Running them individually in an IDE is pretty cool because in your IDE (at least in IntelliJ) you can open the togglz project, make changes to something, and then run just the one benchmark you are interested in to see if your new tweaks made things any better or not. 

I added a baseline one that compares the performance of the togglz vs a boolean if flag, another one that looks at how it performs when you use an activation strategy, and another one that tests how it performs when you use a ScriptEngine.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -189,6 +189,7 @@
     <module>distribution</module>
     <module>cassandra</module>
     <module>hazelcast</module>
+    <module>togglz-benchmarks</module>
   </modules>
 
   <developers>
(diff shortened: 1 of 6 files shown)


chkal:
Awesome! Thank you soooo much for contributing this.

Just a side note. I think I'll rename the module from `togglz-benchmarks` to simply `benchmarks` in the next days as this fits better to the naming scheme of the other modules. Even so it is some kind of "special" module. ;)

Again: Thanks so much for this...

PR #161, opened 2016-05-16
Put FeatureManager instance to cache only if absent

dstepanov:


chkal:
Looks good. Thanks a lot.

PR #163, opened 2016-05-31
System Property Activation Strategy

cpkelley:
Activate a feature based on a system property. The property is string based and is not case sensitive. 

Unit and benchmark tests were also added.


chkal:
Awesome! Thanks a lot! This looks great! Just merged it! :beers: 

One idea: What about making the parameter value optional? So if the users omits the value it would mean: "The feature is active if the system property is set with any value". This way users could simply set something like `-DdevMode` on the command line without the need to set an actual value. So this would be similar to Maven which allows to disable tests by adding `-DskipTests` to the command line. 

Thoughts?

PR #165, opened 2016-06-21
TogglzProperties documentation

ractive:
Makes the properties documentation available in the properties editor in STS and IntelliJ IDEA etc.

See http://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html#configuration-metadata-annotation-processor


marceloverdijk:
Thanks for this @ractive!

PR #166, opened 2016-06-22
Add @FeaturesAreActive annotation for controllers

ractive:
Allows you to easily activate spring MVC `@Controller`s with an annotation like:

``` java
 @Controller
 @RequestMapping("/new-feature")
 @FeaturesAreActive(featureClass=MyFeatures.class, features="NEW_FEATURE")
 public class MyNewFeature {
     @RequestMapping(method = RequestMethod.GET)
     public String newFeature() {
         return ....;
     }

     @FeaturesAreActive(featureClass=MyFeatures.class, features="NEW_SECURE_FEATURE", responseStatus=403)
     @RequestMapping(value="/secure", method = RequestMethod.GET)
     public String newSecureFeature() {
         return ....;
     }
 }
```


chkal:
Hey @ractive,

thanks a lot for the pull request. This sounds very interesting. But could you perhaps clean up the patch a bit. It is very difficult to review. It contains many unrelated changes to the pom.xml files and also code format changes for unrelated code.

Thanks a lot.

ractive:
I splitted the PR in two commits where ther first one (de575aedb9a33904b78bb2ae1ac4bde81ab85d55) just adds all togglz projects to the dependencyManagement so that you can refer to them without having to specify a version.

marceloverdijk:
Hi @ractive this is indeed interesting. Some other people were talking about this earlier as well.

I'm wondering, this `FeatureInterceptor` should it be part of the `spring-boot-starter` module or `spring-core`. This @Conditional 'stuff' is part of Spring Framework itself and has no dependency to Spring Boot, or am I wrong?
In that case it would be better to move the `FeatureInterceptor` to the `spring-core` module and the auto configuration of it in `spring-boot-starter`.

Secondly, this `@FeaturesAreActive` works only at startup when creating the beans, right?
It's important to note that toggling features at runtime will probably not enable/create the beans anymore.
Is this the same for the `@RequestMapping`s? It would be very nice that when you toggle a feature the MVC endpoint becomes (un)available again.

ractive:
> I'm wondering, this FeatureInterceptor should it be part of the spring-boot-starter module or spring-core.

I first thought about putting this into the togglz-spring-web project, but the spring version 3.0.x that is used there is incompatibe with the 3.2 and 4.x line regarding interceptors. So I decided to put it into the spring-boot-starter project which depends on spring boot 1.3.x and spring 4.2.x.
But you should consider bumping the spring version anyway. I guess the 3.0.x versions are not supported anymore anyway and the 3.2.x versions will face EOL at the end of the year.

I could try to bump spring to the latest 3.2.x release and then move the FeatureInterceptor stuff into the togglz-spring-web project. WDYT?

> This `@Conditional` 'stuff' is part of Spring Framework itself and has no dependency to Spring Boot, or am I wrong?

The `@Conditional` stuff if only used in the spring boot auto configuration.

> Secondly, this @FeaturesAreActive works only at startup when creating the beans, right?
The interceptor intercepts every request and the decision if a feature is active (and therefore the controller/method is active) is done at runtime. So toggling a feature on/off will turn the controller on/off at runtime.

The `@RequestMapping` annotations are springmvc annotations and only setup during startup, yes.

ractive:
I bumped the Spring version to 3.2.17 (the latest supported 3.x version) and so I could move the FeatureInterceptor to the togglz-spring-web project - which is nice.

spring-web/src/main/java/org/togglz/spring/web/FeatureInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+import org.togglz.core.context.FeatureContext;
+
+import java.lang.annotation.Annotation;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * This interceptor checks if a controller or controller method is annotaged with the
```
mdomasevicius: Typo - **annotaged**

neocotic:
@ractive Here's what I do in my own projects:

### 1. Create a custom `FeaturesService`

Basically, this is just a Spring bean (generally a Service/Component) that can be used anywhere in your Spring application:

``` java
@Service("features")
@Lazy
public class FeaturesService {

    private final Map<String, Feature> featureMapping = new HashMap<>();

    @PostConstruct
    void initialize() {
        for (Feature feature : FeatureContext.getFeatureManager().getFeatures()) {
            featureMapping.put(feature.name(), feature);
        }
    }

    public boolean isActive(Feature feature) {
        return FeatureContext.getFeatureManager().isActive(feature);
    }

    public boolean isActive(Feature... features) {
        if (features == null || features.length == 0) {
            return false;
        }

        for (Feature feature : features) {
            if (!isActive(feature)) {
                return false;
            }
        }

        return true;
    }

    public boolean isActive(String name) {
        Feature feature = featureMapping.get(name);

        Objects.requireNonNull(feature, () -> "Could not find feature for name: " + name);

        return isActive(feature);
    }

    public boolean isActive(String... names) {
        if (names == null || names.length == 0) {
            return false;
        }

        for (String name : names) {
            if (!isActive(name)) {
                return false;
            }
        }

        return true;
    }
}
```

### 2. Combine `FeaturesService` with Spring Security

You can now use your new `FeaturesService` directly in any Spring Security expression using [Expression-Based Access Control](http://docs.spring.io/spring-security/site/docs/4.2.x/reference/html/el-access.html).

This is perfect for Spring controllers as you can simply use the `@PreAuthorize` annotation on the type and/or request mapping methods:

``` java
@Controller
@RequestMapping("/new-feature")
@PreAuthorize("@features.isActive('NEW_FEATURE')")
public class MyNewFeature {
    
    @RequestMapping(method = RequestMethod.GET)
    public String newFeature() {
        return ...;
    }

    @PreAuthorize("@features.isActive('NEW_SECURE_FEATURE')")
    @RequestMapping(value="/secure", method = RequestMethod.GET)
    public String newSecureFeature() {
        return ...;
    }
}
```

Since you're using the Spring Security mechanism like this; if the feature is inactive, it will reject access to the resource (403). Easy. It will allow you to pass multiple feature names as well, if needed.

I believe that method-specific `@PreAuthorize` always override that of the type instead of augmenting it (e.g. like how `@RequestMapping` works), but I've never found that to be a problem.

### 3. (Optional) Use `FeaturesService` elsewhere

Because `FeaturesService` is a Spring bean, you can inject it anywhere in your code:

``` java
@Service
public class UserService {

   @Resource(name = "features")
   private FeaturesService features;

   public User getUser() {
      if (features.isActive(MyFeatures.NEW_FEATURE)) {
         return ...;
      }

      return ...;
   }
}
```

You can even use Spring Security's [JSP tag library](http://docs.spring.io/spring-security/site/docs/4.2.x/reference/html/taglibs.html):

``` jsp
<%@ taglib prefix="sec" uri="http://www.springframework.org/security/tags" %>

<sec:authorize access="@features.isActive('NEW_FEATURE')">
    ...
</sec:authorize>
```

chkal:
@neocotic Thanks for sharing your solution. I'm sure this will help others too.

@marceloverdijk @ractive What's the state of this PR?

ractive:
@chkal: With fb5cf4aa65e0f495317baf06ddd4577e954bc2e8 I import the spring-boot-dependencies module as a BOM so that the versions are used that are defined in its dependencyManagement.

> What's the state of this PR?

So from my point of view this PR is ready.

chkal:
@ractive Thanks a lot!

Perhaps @marceloverdijk or somebody else with more Spring experience could have a deeper look at this?

marceloverdijk:
I had a quick looks and it seems alright.
Maybe special attention to the dependency declarations as these changed as well.

chkal:
Thanks a lot! I just reviewed the changes and they look good.

I'm not so familiar with Spring Boot, but using their BOM looks like a good idea to me.

I'll merge this now. Thanks again! :beers:

PR #170, opened 2016-06-24
(Fixes #169) - Wrapping FeatureState within CacheEntry so that MemcacheStateReposito…

### LINKED ISSUE #169: MemcacheStateRepository hits datastore multiple times when feature state is not persisted
MemcacheStateRepository should cache null values properly

fabito:
…ry can handle null values properly and avoid hitting datastore

Fixes #169


chkal:
Good catch. Thanks a lot.

fabito:
Hi @chkal , any chance to release a new version with these fixes ?

chkal:
I released 2.3.0.RC2 a few days ago. I would like to release 2.3.0.Final in one or two weeks if no blocker issue comes up. So your fix would be release with 2.3.0.Final.

PR #172, opened 2016-06-28
Address range support in ClientIpActivationStrategy

ractive:


servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
```diff
@@ -39,28 +45,49 @@ public String getName()
...
+         List<CIDRUtils> cidrUtils = new ArrayList<>();
+         for (String part : parts) {
+            try {
+               if (part.contains("/")) {
+                  cidrUtils.add(new CIDRUtils(part));
+               } else {
+                  ips.add(InetAddress.getByName(part));
+               }
+            } catch (Exception e) {
+               log.warn("Ignoring illegal IP address or CIDR range " + part);
```
ractive: Not sure if this exception should be catched here. How would the togglz-console UI handle it?

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
```diff
@@ -39,28 +45,49 @@ public String getName()
...
+               return true;
+            }
+   
+            for (CIDRUtils cidrUtil : cidrUtils) {
+               if (cidrUtil.isInRange(request.getRemoteAddr())) {
+                  return true;
+               }
+            }
+         } catch (UnknownHostException e) {
+            log.error("Illegal address " + request.getRemoteAddr());
```
ractive: Not sure if this exception should be catched here. How would the togglz-console UI handle it?

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
chkal: First you should catch `UnknownHostException` instead of `Exception` here.

And we should perhaps use the `ParameterBuilder` to create a validation rule for parameter values. The builder supports this using regular expressions. But you could also provider you own `Parameter` implementation which provides custom validation code. See `ScriptEngineActivationStrategy` for an example. This way we could catch invalid strategy parameters before they are saved.

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
```diff
@@ -39,28 +45,49 @@ public String getName()
...
+               } else {
+                  ips.add(InetAddress.getByName(part));
+               }
+            } catch (Exception e) {
+               log.warn("Ignoring illegal IP address or CIDR range " + part);
+            }
+         }
+
+         try {
+            if (ips.contains(InetAddress.getByName(request.getRemoteAddr()))) {
```
chkal: Does `InetAddress.getByName()` perform DNS lookups? If so, we should not use it because it will be a performance problem.

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
```diff
@@ -39,28 +45,49 @@ public String getName()
...
 
-         String allowedIpsParam = featureState.getParameter(PARAM_IPS);
-         List<String> allowsIps = Strings.splitAndTrim(allowedIpsParam, "[\\s,]+");
-
-         // TODO: This should support a simple form of subnet matching
-         return allowsIps.contains(request.getRemoteAddr());
-
+         List<String> parts = Strings.splitAndTrim(featureState.getParameter(PARAM_IPS), "[\\s,]+");
+
+         List<InetAddress> ips = new ArrayList<>();
```
chkal: I don't think that you have to build these lists first and validate in a separate code block. Why not doing this in one step. This would even be more efficient, because you have a result as soon as the first entry matches and therefore don't have parse the following addresses.

chkal:
Thanks a lot for providing help with this. I added some comments to your code.

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
ractive: Just checked the implementation. If the given address is an IPv4 or IPv6 address, no DNS lookup is done.

ractive:
I introduced a `Parameter` implementation that does the validation. WDYT?

chkal:
Awesome! Thanks a lot! I'll have a depper look at this ASAP. I'm currently packing my stuff for a two week trip. So it may take some time. Sorry about that.

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
```diff
@@ -39,29 +43,91 @@ public String getName()
...
+               return true;
+            }
+            
+            try {
+               if (part.contains("/")) {
+                  CIDRUtils cidrUtil = new CIDRUtils(part);
+                  if (cidrUtil.isInRange(remoteAddr)) {
+                     return true;
+                  }
+               } else if (InetAddress.getByName(remoteAddr).equals(InetAddress.getByName(part))) {
```
chkal: I think `InetAddress.getByName(remoteAddr)` should be moved out of the loop for performance...

servlet/src/main/java/org/togglz/servlet/activation/ClientIpActivationStrategy.java
```diff
@@ -39,29 +43,91 @@ public String getName()
...
+                  return true;
+               }
+            } catch (Exception e) {
```
chkal: Catching `Exception` is bad practice. Could you catch the corresponding exceptions instead. Catching `Exception` will also catch stuff like NPEs, which is not very nice.

chkal:
Hey. Sorry for the delay. As I wrote earlier I'm currently on vacation and I promised my wife to not use the notebook too much. ;-)

Your code looks great! I just added two comments. Thank you very much for your contribution. This is awesome!

chkal:
And BTW: The Parameter implementation looks great. This way the parameters will be validated correctly and there should never be an invalid IP / netmask, which is great!

ractive:
> I promised my wife to not use the notebook too much. ;-)

Just hold to this promise... ;-)

I implemented your suggestions.

chkal:
Awesome! This looks great! Thank you so much.

But it looks like there are test failures... Not sure why. Actually I don't think it is related to your changes. I'll have to check that...

chkal:
Ok, tests are green now. Seems like this was just a temporary failure. I just merged your changes. Thank you very much! :beers:

ractive:
Nice! I'm looking forward to use this new feature... :-)

chkal:
I hope to find some time at the weekend to cut a release...

chkal:
Hey @ractive,

I've a funny story to share. I just ran into weird test failures with the test you implemented on my notebook. Basically the failing test verified, that `abc/24` isn't a valid IP range. It took me quite some time to figure out that the Google DNS servers which I'm using are able to resolve `abc` as a valid hostname. So the code actually got a valid IP and therefore the range was valid.

    $ dig +short @8.8.8.8 abc
    127.0.53.53

So the current code treats `abc/24` as `127.0.53.53/24`.

I "fixed" this like this: 1609f9b4debec7ee7d71737aca9d252e2e8344b8

I'm not sure if I like that the code currently resolves hostnames. Does this make sense?

ractive:
Of course does *the alphabet* resolve `abc`. :-)

Hm. This "abc/24" was a way to test the IllegalArgumentException that is thrown by the CIDRUtil.
But I agree. We should check that the user only uses IP addresses and not hostnames, esp. for the CIDR range queries. I quickly thought about it, but could not find an easy way to determine if a string is an IP address or not.
I'll give it a try later on...

chkal:
Awesome. Thanks. ;-)

ractive:
Back from holiday. I hope I can give it a shot this week...

PR #181, opened 2016-08-29
Add a state repository that uses Amazon's DynamoDB

ryangardner:
This pull request moves a helper class from the ZookeeperStateRepository to the util package in the core module and adds a state repository that will use Amazon's DynamoDB.

DynamoDB is a natural choice for people who want to use feature toggles on cloud-native applications because it is very simple to create a new one, it's fast, fully managed, and relatively inexpensive. 

A small "README" with the command to use to create the table in dynamo and some sample snippets of how to use it in a spring application are also included. (along with an integration test that will download and run some simple tests using the "DynamoDbLocal" testing app that amazon provides)


chkal:
This looks great! Thank you very much for this contribution! You are awesome! :beers:

PR #182, opened 2016-09-06
Possibility to use all repositories in composite to set the state

hexmind:
It is useful to audit or notify.
Is it OK @chkal  ?


chkal:
Thanks a lot! I like it! Makes sense to me! Merged! ;-)

PR #184, opened 2016-10-17
Device Activation Strategy based on spring mobile

anupriyachhabra:
A new activation strategy to turn on/off a feature based on the device type of user. This is based on spring-mobile project http://projects.spring.io/spring-mobile/. I am currently using this in my work and this looked like a nice thing to give back to the Togglz community.

### DIFF
--- a/spring-mobile/pom.xml
+++ b/spring-mobile/pom.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.togglz</groupId>
+    <artifactId>togglz-project</artifactId>
+    <version>2.4.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>togglz-spring-mobile</artifactId>
+  <name>Togglz - Spring mobile integration</name>
+  <description>Togglz - Spring mobile integration</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.togglz</groupId>
+      <artifactId>togglz-servlet</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.mobile</groupId>
+      <artifactId>spring-mobile-device</artifactId>
+      <version>1.1.4.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.spec.javax.servlet</groupId>
+      <artifactId>jboss-servlet-api_3.0_spec</artifactId>
+      <version>1.0.1.Final</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.togglz</groupId>
+      <artifactId>togglz-test-harness</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+</project>
--- a/spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
+++ b/spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
@@ -1,0 +1,71 @@
+package org.togglz.spring.mobile;
+
+import org.springframework.mobile.device.Device;
+import org.springframework.mobile.device.DeviceType;
+import org.springframework.mobile.device.DeviceUtils;
+import org.togglz.core.activation.Parameter;
+import org.togglz.core.activation.ParameterBuilder;
+import org.togglz.core.repository.FeatureState;
+import org.togglz.core.spi.ActivationStrategy;
+import org.togglz.core.user.FeatureUser;
+import org.togglz.servlet.util.HttpServletRequestHolder;
+
+import javax.servlet.http.HttpServletRequest;
+
+import static org.springframework.mobile.device.DeviceType.MOBILE;
+import static org.springframework.mobile.device.DeviceType.NORMAL;
+import static org.springframework.mobile.device.DeviceType.TABLET;
+
+
+/**
+ * Activation strategy that will use the Device type used by client to decide if a feature is active or not.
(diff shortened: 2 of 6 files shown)


spring-mobile/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.togglz</groupId>
+    <artifactId>togglz-project</artifactId>
+    <version>2.4.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>togglz-spring-mobile</artifactId>
+  <version>2.4.0-SNAPSHOT</version>
```
chkal: You can remove the `<version>` element here as it is inherited from the parent. This way it is easier to do releases.

spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+        return ID;
+    }
+
+    @Override
+    public String getName() {
+        return "Device Rollout";
+    }
+
+    @Override
+    public Parameter[] getParameters() {
```
chkal: A few things in this method:
 * Could you replace the term "Test" with "Feature" in the description? I think this makes more sense
 * What do you think about also allowing lowercase "yes" for activating the feature? Should work by addding `(?i)` to the regex. And then simply compare using `equalsIgnoreCase()`
 * Using the key `ON+DESKTOP` looks weird. You can remove the `ON` part and just use simple lower cases keys like `mobile` or `desktop`. That's like the other strategies work. I think it would be great to be consistent here.

spring-mobile/src/test/java/org/togglz/spring/mobile/DeviceActivationStrategyTest.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.togglz.spring.mobile.DeviceActivationStrategyTest.MockRequestAssert.assertThat;
+import static org.togglz.spring.mobile.DeviceActivationStrategy.DESKTOP;
+import static org.togglz.spring.mobile.DeviceActivationStrategy.TABLET;
+import static org.togglz.spring.mobile.DeviceActivationStrategy.MOBILE;
+
+
+/**
+ * Created by achhabra on 10/17/16.
+ */
+public class DeviceActivationStrategyTest {
```
chkal: Great test coverage! :+1:

spring-mobile/template.mf
```diff
@@ -0,0 +1,12 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-SymbolicName: org.togglz.spring.mobile
+Bundle-Vendor: http://www.togglz.org/
+Bundle-Version: ${project.version}
+Bundle-Name: ${project.name}
+Version-Patterns:
+ default;pattern="[=.=.=, =.+1.0)"
+Import-Template:
```
chkal: Looks fine to me. As I wrote before: I'm not so familiar with OSGi myself, but the Maven plugin will fail if there are any packages missing. So if the build is green everything is fine.

spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
hexmind: I strongly agree with prefix removal. What is more first 3 constants are good enum candidate. You can also reuse existing one: org.springframework.mobile.device.DeviceType.

spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+        return new Parameter[]{
+                ParameterBuilder.create(ON+DESKTOP).label("Turn this test ON for desktop").matching("(YES|NO)")
```
hexmind: Can you break lines before .label, .matching and .description parts? It will be more readable.

spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+                        "Test will be off by default enter 'YES' in box above to enable")
+        };
+    }
+
+    @Override
+    public boolean isActive(FeatureState featureState, FeatureUser user) {
+
+        HttpServletRequest request = HttpServletRequestHolder.get();
+        if (request != null) {
+            Device device = DeviceUtils.getCurrentDevice(request);
```
hexmind: Here is main recognition logic. I suggest to extract private method
    String getParameterName(HttpServletRequest request)
and give up with many ternary operators in one line.

spring-mobile/src/test/java/org/togglz/spring/mobile/DeviceActivationStrategyTest.java
hexmind: I suggest extract MockRequest and MockRequestAssert to package scope classes (separated files).

anupriyachhabra: @hexmind dont want to extract MockRequest to separate class as it contains logic for mocks specific to this strategy only

spring-mobile/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.spec.javax.servlet</groupId>
+      <artifactId>jboss-servlet-api_3.0_spec</artifactId>
+      <version>1.0.1.Final</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
```
anupriyachhabra: done

anupriyachhabra:
@chkal and @hexmind I have updated the Pull Request with the changes suggested in review comments. Can you review again?

Thanks

spring-mobile/src/main/java/org/togglz/spring/mobile/DeviceActivationStrategy.java
anupriyachhabra: @hexmind I am a groovy dev and really like ternary operators and reducing lines of code using them.

hexmind: Security issue? No. getParameters must be public, but getParameterName should be private.

chkal:
Awesome! Thanks a lot for applying the suggested changes! I just merged your changes!

PR #186, opened 2016-11-25
Spring environment properties activation strategy

neocotic:
This adds a new activation strategy designed to support Environment properties to integrate feature toggle properties into the same property sources as the rest of the application properties.

It supports a parameter to specify the property but will gracefully fallback to a property name generated from the `Feature` instance and its class (e.g. `com.example.MyFeatures.FEATURE_NAME`).

Related Spring Docs: http://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-property-source-abstraction

### DIFF
--- a/spring-core/pom.xml
+++ b/spring-core/pom.xml
@@ -47,16 +47,5 @@
       <scope>test</scope>
     </dependency>
 
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-
   </dependencies>
 </project>
--- a/spring-core/pom.xml
+++ b/spring-core/pom.xml
@@ -13,7 +13,7 @@
   <description>Togglz - Spring Framework integration</description>
 
   <properties>
-    <spring.version>3.0.7.RELEASE</spring.version>
+    <spring.version>3.1.4.RELEASE</spring.version>
   </properties>
 
   <dependencies>
@@ -28,6 +28,23 @@
       <artifactId>spring-context</artifactId>
       <version>${spring.version}</version>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+      <version>${spring.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
     </dependency>
 
   </dependencies>
--- a/spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
+++ b/spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
@@ -51,7 +51,8 @@ public class SpringEnvironmentPropertyActivationStrategy implements ActivationSt
     public boolean isActive(FeatureState featureState, FeatureUser user) {
         ApplicationContext applicationContext = ContextClassLoaderApplicationContextHolder.get();
         if (applicationContext == null) {
-            return false;
+            throw new IllegalStateException("ApplicationContext could not be found, which can occur if there is no bean for "
+                + "TogglzApplicationContextBinderApplicationListener when TogglzAutoConfiguration is not being used");
         }
 
         Environment environment = applicationContext.getEnvironment();
--- a/spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
+++ b/spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
@@ -25,7 +25,6 @@ import org.togglz.spring.util.ContextClassLoaderApplicationContextHolder;
 public class SpringEnvironmentPropertyActivationStrategy implements ActivationStrategy {
 
     public static final String ID = "spring-environment-property";
-    static final String NAME = "Spring Environment Property";
     public static final String PARAM_NAME = "name";
 
     private static String getPropertyName(FeatureState featureState) {
@@ -45,7 +44,7 @@ public class SpringEnvironmentPropertyActivationStrategy implements ActivationSt
 
     @Override
     public String getName() {
-        return NAME;
+        return "Spring Environment Property";
     }
 
     @Override
--- a/spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
+++ b/spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
@@ -1,0 +1,71 @@
+package org.togglz.spring.activation;
+
+import org.springframework.context.ApplicationContext;
+import org.springframework.core.env.Environment;
+import org.togglz.core.Feature;
+import org.togglz.core.activation.Parameter;
+import org.togglz.core.activation.ParameterBuilder;
+import org.togglz.core.repository.FeatureState;
+import org.togglz.core.spi.ActivationStrategy;
+import org.togglz.core.user.FeatureUser;
+import org.togglz.core.util.Strings;
+import org.togglz.spring.util.ContextClassLoaderApplicationContextHolder;
+
+/**
+ * <p>
+ * An activation strategy based on the values of properties accessible within the Spring environment.
+ * </p>
+ * <p>
+ * It can either be based on a given property name, passed as a parameter, or a property name constructed from the
+ * {@link Feature} itself (e.g. {@code com.example.MyFeatures.FEATURE_NAME}).
+ * </p>
+ *
+ * @author Alasdair Mercer
+ */
+public class SpringEnvironmentPropertyActivationStrategy implements ActivationStrategy {
+
+    public static final String ID = "spring-environment-property";
+    static final String NAME = "Spring Environment Property";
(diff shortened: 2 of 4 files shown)


chkal:
Awesome! Thanks a lot for providing this pull request. I'll review it in detail ASAP and come back to you after that.

neocotic:
@chkal Great! I have an idea for another Spring activation strategy but I'll hold off on that as they'll be a bit of a crossover and don't want to duplicate any issues that you'd like me to correct here.

I'm happy to make any changes required. Thanks for the great library.

chkal:
Awesome! Thanks for your patience. I'm very busy at the moment, so it may take some time. Anyway, thanks a lot for your contribution!

BTW: Feel free to join the Gitter channel. This is a great place to start discussing ideas. ;-)

https://gitter.im/togglz/togglz

spring-core/pom.xml
```diff
@@ -13,7 +13,7 @@
   <description>Togglz - Spring Framework integration</description>
 
   <properties>
-    <spring.version>3.0.7.RELEASE</spring.version>
+    <spring.version>3.1.4.RELEASE</spring.version>
```
neocotic: The `Environment` class was introduced in v3.1. I've imported 3.1.4 as this was the latest fix version of 3.1.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>
+ * It can either be based on a given property name, passed as a parameter, or a property name constructed from the
+ * {@link Feature} itself (e.g. {@code com.example.MyFeatures.FEATURE_NAME}).
+ * </p>
+ *
+ * @author Alasdair Mercer
+ */
+public class SpringEnvironmentPropertyActivationStrategy implements ActivationStrategy {
+
+    public static final String ID = "spring-environment-property";
```
neocotic: I was simply following the standard of having ID and PARAM constants public so that they can be used by consumers of the API, presumably to avoid them from hard-coding these values and potentially breaking on upgrades.

spring-core/src/test/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategyTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+        assertNotNull(parameter);
+        assertEquals(SpringEnvironmentPropertyActivationStrategy.PARAM_NAME, parameter.getName());
+        assertTrue(parameter.isOptional());
+        assertTrue(Strings.isNotBlank(parameter.getLabel()));
+        assertTrue(Strings.isNotBlank(parameter.getDescription()));
+    }
+
+    public enum TestFeatures implements Feature {
+
```
neocotic: This was formatted using the Eclipse settings included in the repository. It didn't appear to mind this.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public boolean isActive(FeatureState featureState, FeatureUser user) {
+        ApplicationContext applicationContext = ContextClassLoaderApplicationContextHolder.get();
+        if (applicationContext == null) {
+            return false;
```
neocotic: I based this on how existing activation strategies handled this. Specifically, I tried to mimic the conditional handling of `HttpServletRequest` within the existing `DeviceActivationStrategy`, `ClientIpActivationStrategy`, and `ServerNameActivationStrategy` classes.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        Environment environment = applicationContext.getEnvironment();
```
neocotic: I'm not sure that I agree and it seems to contradict what I'm seeing in the code base. A good example is the existing `ClientIpActivationStrategy` class which was created by the Togglz author and makes good use of multiple return statements. When used correctly, I believe they are a good tool for avoiding heavy block nesting which becomes even harder to read and follow.

spring-core/pom.xml
chkal: I'm basically fine with increasing the version to 3.1.4.Final. I just did some research and it looks like even Spring 3.2.x will be EOLed by the end of this year. So moving to 3.1.x is fine.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
chkal: I'm fine with using `public` here. This may especially be useful if users want to update the feature state problematically.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
chkal: Hmmm. Interesting question. Back then I handled the missing `HttpServletRequest` very gracefully because I didn't want to throw an exception if the user for some reason uses the toggle outside of a web request (like in a Quartz job for example).

However, we could consider a missing ApplicationContext in a Spring application as a major problem because in this case there is typically something wrong with your setup. So throwing an exception would be OK here.

spring-core/src/test/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategyTest.java
chkal: Maybe the code formatter I created isn't so optimal. I'm not using Eclipse any more these days, so I didn't notice. But this isn't a blocker.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
chkal: I'm also -1 for adding an else here. But that's a matter of taste.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
neocotic: No problem. I'll look at making this change here and in #188 as well. Any preferred exception type? I'm currently thinking `IllegalStateException`.

chkal: `IllegalStateException` with some which guides the user into the right direction would be awesome.

spring-core/src/main/java/org/togglz/spring/activation/SpringEnvironmentPropertyActivationStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+    public static final String PARAM_NAME = "name";
+
+    private static String getPropertyName(FeatureState featureState) {
+        String propertyName = featureState.getParameter(PARAM_NAME);
+        if (Strings.isNotBlank(propertyName)) {
+            return propertyName;
+        }
+
+        Feature feature = featureState.getFeature();
+        return feature.getClass().getName() + "." + feature.name();
```
hexmind: My mistake, please ignore it.

neocotic:
I've committed the recommended changes (i.e. `NAME` constant moved to inline literal and `isActive` method throws when `ApplicationContext` was not found).

chkal:
That's awesome! I just reviewed everything and it looks great!

There is just one thing I'm not completely sure about. It is this line:

    Feature feature = featureState.getFeature();
    return feature.getClass().getName() + "." + feature.name();

So if I got everything correctly, this is basically a fallback case, if the user didn't specify an environment property name. I like the idea of providing a fallback here. That's basically a good example for CoC.

However, I don't think that using `featureState.getFeature().getClass().getName()` as a prefix will work in all cases for several reasons:

  * Togglz allows to use other ways of defining features than enums. You can for example define which features your app supports with properties files. In this case the implementations of `Feature` will be an Togglz internal class: https://github.com/togglz/togglz/blob/master/core/src/main/java/org/togglz/core/util/NamedFeature.java
  * IIRC there are cases where `FeatureManager.isActive(Feature)` gets called with a `NamedFeature` instead of a real enum value. I don't remember all the details any more, but this may lead to unexpected behavior if this code is used.

I basically understand that prefixing is a good idea for creating "namespaces", but what do you think about removing the prefix? Or using something like `togglz.` as a prefix?

BTW: The state repositories also use `Feature.name()` as a key without any prefix for storing the features state.

chkal:
BTW: See this class for the an example for how to specify feature without enums:

https://github.com/togglz/togglz/blob/master/core/src/main/java/org/togglz/core/manager/PropertyFeatureProvider.java

neocotic:
I really wanted to make the property name to be optional so I'm glad you like that aspect of the implementation.

I did really like the fact that the default property name was based on the class name of the feature but, if that doesn't always make sense to use, I will happily use something more generic. The main reason that I used the class name was to avoid possible conflicts as I was under the impression that multiple `Feature` enums could be configured as `EnumBasedFeatureProvider` can take multiple enums. I guess in such cases the developer would simply have to specify a property to avoid conflicts, I could mention this in the JavaDoc for the class.

I'll make the change to switch the prefix with "togglz." as I think we still need a prefix to avoid conflicts with properties for other frameworks/libraries used by the application.

On a minor note; what do you think of the naming of this class (and the one for #188)? Do you think that the "Spring" class name prefix is unnecessary or too granular?

neocotic:
I've committed the changes relating to the prefix.

chkal:
Hey! Awesome! Thanks a lot for applying this change so quickly. I agree that we need the `togglz.` prefix to prevent name collisions with other frameworks.

I'm fine with the Spring prefix in the class name. Lets keep this for now.

Thank you so much for your work on this. I'll merge the PR now. :beers:

PR #188, opened 2016-11-28
Spring profile activation strategy

neocotic:
This adds a new activation strategy designed to allow features to be activated depending on the active profiles within the Spring Environment.

Related Spring Docs: http://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-definition-profiles

### DIFF
--- a/spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
+++ b/spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
@@ -37,7 +37,8 @@ public class SpringProfileActivationStrategy implements ActivationStrategy {
     public boolean isActive(FeatureState featureState, FeatureUser user) {
         ApplicationContext applicationContext = ContextClassLoaderApplicationContextHolder.get();
         if (applicationContext == null) {
-            return false;
+            throw new IllegalStateException("ApplicationContext could not be found, which can occur if there is no bean for "
+                + "TogglzApplicationContextBinderApplicationListener when TogglzAutoConfiguration is not being used");
         }
 
         List<String> activeProfiles = Arrays.asList(applicationContext.getEnvironment().getActiveProfiles());
--- a/spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
+++ b/spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
@@ -21,7 +21,6 @@ import org.togglz.spring.util.ContextClassLoaderApplicationContextHolder;
 public class SpringProfileActivationStrategy implements ActivationStrategy {
 
     public static final String ID = "spring-profile";
-    static final String NAME = "Spring Profile";
     public static final String PARAM_PROFILES = "profiles";
 
     @Override
@@ -31,7 +30,7 @@ public class SpringProfileActivationStrategy implements ActivationStrategy {
 
     @Override
     public String getName() {
-        return NAME;
+        return "Spring Profile";
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


neocotic:
There's some overlap between this and #186 in terms of implementation but I wanted to get them both submitted before I forgot.

spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>
+ * An activation strategy based on the profiles that are active within the Spring environment.
+ * </p>
+ *
+ * @author Alasdair Mercer
+ */
+public class SpringProfileActivationStrategy implements ActivationStrategy {
+
+    public static final String ID = "spring-profile";
+    static final String NAME = "Spring Profile";
```
chkal: I suggest to either make this private or to inline the field. Letting getName() return a literal is fine for me. ;-)

spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public boolean isActive(FeatureState featureState, FeatureUser user) {
+        ApplicationContext applicationContext = ContextClassLoaderApplicationContextHolder.get();
+        if (applicationContext == null) {
+            return false;
```
chkal: We COULD think about treating this as a fatal error. Not 100% sure about this. See my thoughts in the other PR.

spring-core/src/main/java/org/togglz/spring/activation/SpringProfileActivationStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        List<String> activeProfiles = Arrays.asList(applicationContext.getEnvironment().getActiveProfiles());
+        List<String> profiles = Strings.splitAndTrim(featureState.getParameter(PARAM_PROFILES), "[\\s,]+");
+
+        for (String profile : profiles) {
+            if (activeProfiles.contains(profile)) {
```
chkal: Should we perhaps make this check case insensitive? I'm +/-0 on this. Just a thought.

neocotic:
@chkal I've just committed the changes to address your comments. This includes making the profile name checking case-insensitive. Although I couldn't find any official Spring documentation which states that they are not case-sensitive, I did find quite a lot of articles stating that Spring ignores case for profile names.

I've also expanded on the original functionality a bit to better match how profiles are generally used in Spring applications by supporting the NOT operator (`!`). You'd generally see profiles used like this in Spring beans:

``` java
@Configuration
@Profile("production")
public class ProductionConfiguration {
    // ...
}
```

The above would result in `ProductionConfiguration` only being registered if the "production" profile was active. Simple.

But Spring also allows you to register beans only when a profile is **not** active, like the following:

``` java
@Configuration
@Profile("!production")
public class DevelopmentConfiguration {
    // ...
}
```

This time the above would result in `DevelopmentConfiguration` only being registered if the "production" profile was inactive.

The same can now be done for features :smile:

``` java
public enum MyFeatures implements Feature {

    @EnabledByDefault
    @DefaultActivationStrategy(
        id = SpringProfileActivationStrategy.ID,
        parameters = {
            @ActivationParameter(name = SpringProfileActivationStrategy.PARAM_PROFILES, value = "!production")
        }
    )
    FEATURE_ONE
}
```

The code still supports a comma-separated list of profile names as the parameter value and the feature will be active if any of the profiles (possibly negated) match up with those that are currently active. For example; If the value of the parameter was`"p1,!p2"`, the feature would only be active if "p1" is active or if "p2" is not active.

chkal:
Awesome! Looks great! I love the "negate" feature! It totally makes sense.

I just wanted to merge this, but it looks like we are getting merge conflicts because this PR contains some changes which also were contained in the other PR (like the Spring version update). 

Could you perhaps rebase this one against the current master branch? Then I can merge it right away.

Thanks! :beers:

neocotic:
I merged in the latest changes from `master` and it looks like there's no conflicts anymore.

chkal:
Thank you so much! Merged!

PR #189, opened 2016-11-29
Moved TogglzApplicationContextBinderApplicationListener to togglz-spring-core

neocotic:
The Spring `BeanFinder` implementations rely on `ContextClassLoaderApplicationContextHolder` to provide the required `ApplicationContext`. However, currently, the only time that `ContextClassLoaderApplicationContextHolder#bind` is called with the `ApplicationContext` instance is from within `TogglzApplicationContextBinderApplicationListener` which resides within togglz-spring-boot-starter.

This isn't a huge issue as applications can simply depend on that module and have access to it. However, it feels strange that I'm having to depend on a module targeting Spring Boot to make something work that's in togglz-spring-core. This PR simply moves the `TogglzApplicationContextBinderApplicationListener` class from togglz-spring-boot-starter into togglz-spring-core. This means that applications not really caring about the Spring Boot automatic configuration etc can still register this bean in order to make full use of togglz-spring-core using something like the following:

Java-based Spring configurations:
``` java
@Bean
public TogglzApplicationContextBinderApplicationListener togglzApplicationContextBinderApplicationListener() {
    return new TogglzApplicationContextBinderApplicationListener();
}
```

XML-based Spring configurations:
``` xml
<bean name="togglzApplicationContextBinderApplicationListener" class="org.togglz.spring.listener.TogglzApplicationContextBinderApplicationListener" />
```

I must state that this would obviously be a breaking change for anyone already doing something like the above as the package will have changed and they may need to add the togglz-spring-core dependency. This would be something best captured on the [Updating Notes](https://www.togglz.org/updating-notes.html) page.

Currently, the `ContextClassLoaderApplicationContextHolder` is only really being used by the Spring `BeanFinder` implementations. However, both of the Spring-related `ActivationStrategy` implementations that I've submitted (#186 & #188) rely on it too in order to get the Spring `Environment` instance that they require. If this listener is not registered as a bean, either explicitly or via the Spring Boot auto-configuration, those activation strategies will never activate features as they'll fail to lookup the `ApplicationContext`.

### DIFF
--- a/spring-boot-starter/src/main/java/org/togglz/spring/boot/autoconfigure/TogglzAutoConfiguration.java
+++ b/spring-boot-starter/src/main/java/org/togglz/spring/boot/autoconfigure/TogglzAutoConfiguration.java
@@ -41,7 +41,6 @@ import org.togglz.core.manager.EmptyFeatureProvider;
 import org.togglz.core.manager.EnumBasedFeatureProvider;
 import org.togglz.core.manager.FeatureManager;
 import org.togglz.core.manager.FeatureManagerBuilder;
-import org.togglz.core.metadata.FeatureMetaData;
 import org.togglz.core.repository.StateRepository;
 import org.togglz.core.repository.cache.CachingStateRepository;
 import org.togglz.core.repository.composite.CompositeStateRepository;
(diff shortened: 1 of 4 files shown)


chkal:
Awesome. Thanks a lot for providing the detailed explanation.

I must confess that I'm not so familiar with Spring and the recent developments like Spring Boot. Perhaps @marceloverdijk could tell us if there is anything that blocks moving this class to the core package!?!

Regarding compatibility: I'm fine with adding a note to the "Updating Notes". This should be fine.

Thanks a lot.

marceloverdijk:
I will try to have a look this week!

chkal:
Awesome! Great! So I'll merge this!

chkal:
Merged! I'll leave this open so we don't forget to add a note to the "Updating Notes" page.

@neocotic You can BTW also provide PRs for the Togglz site:

https://github.com/togglz/togglz-site/blob/master/updating-notes.html.haml

PR #191, opened 2016-11-30
Introducing MongoStateRepository for Mongo 3 driver

kevinstembridge:


chkal:
Awesome! Thanks a lot. I'll have a deeper look at this ASAP.

But one first thought: I don't think we should create a separate module for this? Why not just updating the driver of the standard `togglz-mongodb` module? Or are there many people who keep using the 2.x version for some reason?

kevinstembridge:
It would be nice to avoid having 2 modules but that would be a breaking change for anyone who needs the 2.x version. There are often multiple libraries in the same JVM using the Mongo driver, so flexibility would be a good thing. Ultimately, it's your call though. Whatever you think is best.

BTW, I tried to update the documentation but got an unhelpful error trying to install awestruct. Haven't tried to go any further with it.

kevinstembridge:
I managed to get awestruct working. (needed to update my version of Ruby). 

I'll wait to see what the finished product looks like before I write up some documentation though.

kevinstembridge:
No worries. I'll update the pull request, replacing Mongo 2 with 3.

chkal:
Awesome! This looks great! Thank you so much.

I guess the only thing missing is a way for the developer to specify the `MongoCredential`s to use. But I could take care of this after we merged this.

Thoughts?

kevinstembridge:
I had a quick look at that MongoCredentials stuff and I'm pretty sure it all applies at the level of the MongoClient. If that's the case then I don't think there will be much demand to be able to specify credentials to the MongoStateRepository builder. I think almost everyone who uses the MongoStateRepository will be using Mongo as their application database and will  be sharing the same MongoClient instance that they use within the application. 

Also, I think that the most common way to specify authentication credentials is to provide username and password in the MongoClientURI (works the same as a JDBC connection string). That is then used as input to create the MongoClient.

I'm pretty sure that accepting a MongoClient instance in the repository builder will satisfy just about everyone. But over to you to make the call. :)

chkal:
Okay, great! So I'll merge this one now! Thanks a lot for your help! :beers:

PR #193, opened 2016-12-19
FeatureGroup extends Serializable

bakito:
In my project i want to serialize feature groups.

### DIFF
--- a/core/src/main/java/org/togglz/core/metadata/FeatureGroup.java
+++ b/core/src/main/java/org/togglz/core/metadata/FeatureGroup.java
@@ -1,5 +1,7 @@
 package org.togglz.core.metadata;
 
+import java.io.Serializable;
+
 import org.togglz.core.Feature;
 
 /**
@@ -9,7 +11,7 @@ import org.togglz.core.Feature;
  * @author Christian Kaltepoth
  * 
  */
-public interface FeatureGroup {
+public interface FeatureGroup extends Serializable {
 
     String getLabel();
 
--- a/core/src/main/java/org/togglz/core/metadata/FeatureGroup.java
+++ b/core/src/main/java/org/togglz/core/metadata/FeatureGroup.java
@@ -1,7 +1,5 @@
 package org.togglz.core.metadata;
 
-import java.io.Serializable;
-
 import org.togglz.core.Feature;
 
 /**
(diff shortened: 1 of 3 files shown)


chkal:
Hey. Thanks a lot for the PR. It looks good. But I would like to know why you have the need to serialize feature groups. That's surprising to me. What is the use case?

bakito:
Hi

I'm working on an client server application, where I want so send the features from the server to the client.
I'm working with an EnumBasedFeatureProvider and annotation Groups.
Since I want to send all the FeatureMetaData with the groups to the client, it would be nice if the groups where serializable.
I was not able to restor the AnnotationFeatureGroups on the client side.

chkal:
Now I'm curious. Do you Java serialization for that?

bakito:
Yes the FeatureMetaData is serialized from the server to the client. Currently I buildt some Custom Implementations of FetureGroup. But like this I'm loosing the AnnotationFeatureGroups .

chkal:
To be honest, I'm not sure whether serialization is the best way to do this. However, I cannot really judge about this without knowing more details.

About your PR. I'm basically fine with making `AnnotationFeatureGroup` and `SimpleFeatureGroup` serializable. But I don't like to let the `FeatureGroup` interface itself extend `Serializable`, because this forces ALL implementations (current and future ones) to support serialization. So I would prefer to don't touch `FeatureGroup` and just add the serialization support to the two implementations.

bakito:
I've removed Serializable from the interafce.

The communication between my server and client works with EJB and serialization. I can not change this.
Do you have a better idea on sending the information to the client via serialization?

chkal:
Thanks for reverting the change of the interfaces. But now you will have to add `implements Serializable` to the implementations or it won't work.

Regarding the use of serialization: I don't know much about your use case, but I don't fully understand why the FeatureGroup implementation has to be serialized. I'm sure there are other options. If you really want to manage the feature state on the server, you could for example create a custom state repository which delegates the feature state reading and writing to the server. In this case you will just serialize the FeatureState, which is already serializable.

bakito:
After some Trial anderen error I managed to send the Features to the Client, without serializing the groups.

As you mentioned, I User a custom Statement Report.
Thanks för the advice.

PR #194, opened 2016-12-24
release date activation strategy: show current date in format examples

thokrae:
Maybe showing the current date and time in the examples for the release date activation strategy simplifies the configuration for some users. At least I am always unsure wether it's UTC or local time.

### DIFF
--- a/core/src/main/java/org/togglz/core/activation/ReleaseDateActivationStrategy.java
+++ b/core/src/main/java/org/togglz/core/activation/ReleaseDateActivationStrategy.java
@@ -1,11 +1,9 @@
 package org.togglz.core.activation;
 
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
+import java.text.*;
 import java.util.Date;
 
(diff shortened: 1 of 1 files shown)


chkal:
Hey @thokrae,

thanks a lot for the pull request. But actually I'm not sure if showing the current date/time is a very good idea. There is a specific reason why the help text contains `2012-12-31`: This will tell the user which segment contains the day and which segment contains the month part of the date. I think this is very important especially for users which use different date formats. Which you patch applied the text could display something like `2016-01-01` which is a bad example for showing the user where the day and month is located.

However, if you feel that the missing time zone information may be confusing, we could add this to the help text. What do you think?

thokrae:
Hello @chkal,

thanks for checking the pull request.

Good point about the date format. You are right, extending the help text is a better solution for this issue.

chkal:
I'll close this one now. Pull requests are welcome if there is a need to optimize the help text.

PR #195, opened 2016-12-27
Add Amazon s3 implementation

MarkRx:
Adds amazon s3 StateRepository implementation. Each feature is stored as json in the specified bucket with a key of togglz/<feature_name> by default. I followed the pattern that other implementations used.


MarkRx:
Need to review first.

chkal:
Hey, thanks. I'm just wondering why someone would prefer S3 over DynamoDB on AWS? Any insights?

MarkRx:
DynamoDB may not be available. An S3 layer can actually exist inside an internal cloud using a product with an S3 API facade such as Dell ECS (https://www.emc.com/techpubs/ecs/ecs_s3_supported_features-1.htm).

This was part of a POC for using S3 to store feature switch information. We're still discussing what the best place to store that information is though (such as mongo) so if you have any suggestions that would be appreciated.

MarkRx:
All good.

MarkRx:
done

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+ * 
+ * @author Mark Richardson
+ * @date 12/27/16
+ */
+public class S3StateRepository implements StateRepository {
+
+    // http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
+    private static final String ERR_NO_SUCH_KEY = "NoSuchKey";
+
+    private ObjectMapper objectMapper = new ObjectMapper();
```
chkal: I think this could get `final` too....

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+            if (object != null) {
+                FeatureStateStorageWrapper wrapper = objectMapper.reader().forType(FeatureStateStorageWrapper.class).readValue(object.getObjectContent());
+                return FeatureStateStorageWrapper.featureStateForWrapper(feature, wrapper);
+            }
+        } catch (AmazonS3Exception ae) {
+            if (ERR_NO_SUCH_KEY.equals(ae.getErrorCode())) {
+                return null;
+            }
+            throw ae;
+        } catch (Exception e) {
```
chkal: Catching `Exception` is commonly considered bad practice. So I would prefer to either remove this block or to catch more specific exception types.

chkal:
> DynamoDB may not be available. An S3 layer can actually exist inside an internal cloud using a product with an S3 API facade such as Dell ECS (https://www.emc.com/techpubs/ecs/ecs_s3_supported_features-1.htm).

Ah, OK... So you are talking about the case that there is just a S3 compatible interface and not the read AWS infrastructure available.

> This was part of a POC for using S3 to store feature switch information. We're still discussing what the best place to store that information is though (such as mongo) so if you have any suggestions that would be appreciated.

I'm still not completely sure if it makes sense to use S3 for that. In a cluster environment you typically have some other shared data store available which may be a better fit. A relational database, a key-value store, or something else. Using S3 which is an object store looks weird to me. But I may be wrong about that.

ryangardner:
I can see some benefits of S3 over dynamo. In both cases you are likely
going to use a cache in front of it to avoid the latency of hitting the
store each time (and in dynamo's case to help keep the costs down since you
pay for read capacity)

With S3 you could still use the cache layer in front of the repository but
the costs could be cheaper. (Though we are talking about potentially saving
only a couple of dollars a month so it's not like it's a massive savings.)

I haven't looked over the code in much detail but for some cases S3 might
be a better fit than dynamo.

chkal:
Thanks @ryangardner for sharing your thoughts. I agree that using a cache layer is a good choice in most scenarios. Especially if you pay per request to the underlying data store.

pom.xml
```diff
@@ -187,6 +187,7 @@
     <module>hazelcast</module>
     <module>zookeeper</module>
     <module>dynamodb</module>
+    <module>s3</module>
```
hexmind: I suggest rename to amazon-s3. It will be still short. Today S3 abbreviation is obvious, but in future who knows.

s3/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+
+		<dependency>
+			<groupId>com.amazonaws</groupId>
+			<artifactId>aws-java-sdk-s3</artifactId>
+			<version>1.11.29</version>
+		</dependency>
+
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
```
hexmind: `<scope>test</scope>`

s3/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+		</dependency>
+
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
```
hexmind: `<scope>test</scope>`

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+    // http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
+    private static final String ERR_NO_SUCH_KEY = "NoSuchKey";
+
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    protected final AmazonS3Client client;
+    protected final String bucketName;
+    protected final String keyPrefix;
+
+    private S3StateRepository(Builder builder) {
```
hexmind: Builder use should be optional. It is more typical approach for example with dependency injection.
This constructor should accept all fields as arguemnts, like in builder::build:
`return new S3StateRepository(client, bucketName, keyPrefix);`

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    public FeatureState getFeatureState(Feature feature) {
+        try (S3Object object = client.getObject(bucketName, keyPrefix + feature.name())) {
+            if (object != null) {
+                FeatureStateStorageWrapper wrapper = objectMapper.reader().forType(FeatureStateStorageWrapper.class).readValue(object.getObjectContent());
+                return FeatureStateStorageWrapper.featureStateForWrapper(feature, wrapper);
+            }
+        } catch (AmazonS3Exception ae) {
+            if (ERR_NO_SUCH_KEY.equals(ae.getErrorCode())) {
+                return null;
```
hexmind: please add log.warn()

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+
+        return null;
+    }
+
+    @Override
+    public void setFeatureState(FeatureState featureState) {
+        try {
+            String json = objectMapper.writeValueAsString(FeatureStateStorageWrapper.wrapperForFeatureState(featureState));
+            client.putObject(bucketName, keyPrefix + featureState.getFeature().name(), json);
+        } catch (Exception e) {
```
hexmind: I would prefer to catch more specific exception type too.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
hexmind: You could use typed, custom exception:
throw new TogglzS3Exception("Cannot read feature state", e);
throw new TogglzS3Exception("Cannot write feature state", e);

s3/src/main/java/org/togglz/s3/S3StateRepository.java
hexmind: please extract local var:
`String json = objectMapper.writeValueAsString(storageWrapper);`

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return a Builder
+     */
+    public static Builder newBuilder(AmazonS3Client client, String bucketName) {
+        return new Builder(client, bucketName);
+    }
+
+    /**
+     * Builder for a {@link S3StateRepository}.
+     */
+    public static class Builder {
```
hexmind: Please move it to package-scoped class S3StateRepositoryBuilder.

s3/src/test/java/org/togglz/s3/S3StateRepositoryTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.Bucket;
+import com.amazonaws.services.s3.model.PutObjectResult;
+import com.amazonaws.services.s3.model.S3Object;
+import com.amazonaws.services.s3.model.S3ObjectInputStream;
+
+public class S3StateRepositoryTest {
+    @SuppressWarnings("serial")
+    @Test
+    public void test() {
```
hexmind: Please divide this test to more cases and name each method.

s3/src/test/java/org/togglz/s3/S3StateRepositoryTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertNull(repository.getFeatureState(TestFeature.FEATURE_1));
+
+        FeatureState initFs = new FeatureState(TestFeature.FEATURE_1)
+            .setEnabled(true)
+            .setStrategyId("abc")
+            .setParameter("key1", "value1");
+
+        repository.setFeatureState(initFs);
+
+        FeatureState actualFs = repository.getFeatureState(TestFeature.FEATURE_1);
```
hexmind: actualState

s3/src/test/java/org/togglz/s3/S3StateRepositoryTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void test() {
+        AmazonS3Client client = new AmazonS3ClientMOCK(new AnonymousAWSCredentials());
+        client.withEndpoint(String.format("http://localhost:8001"));
+        client.createBucket("testbucket");
+
+        S3StateRepository repository = S3StateRepository.newBuilder(client, "testbucket").build();
+
+        assertNull(repository.getFeatureState(TestFeature.FEATURE_1));
+
+        FeatureState initFs = new FeatureState(TestFeature.FEATURE_1)
```
hexmind: initState

s3/src/test/java/org/togglz/s3/S3StateRepositoryTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertThat(actualFs.getStrategyId()).isEqualTo("abc");
+        assertThat(actualFs.isEnabled()).isEqualTo(true);
+        assertThat(actualFs.getParameter("key1")).isEqualTo("value1");
+        assertThat(actualFs.getParameterNames()).isEqualTo(new HashSet<String>() {
+            {
+                add("key1");
+            }
+        });
+
+        FeatureState updatedFs = new FeatureState(TestFeature.FEATURE_1)
```
hexmind: updatedState

s3/src/test/java/org/togglz/s3/S3StateRepositoryTest.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+        });
+    }
+
+    private enum TestFeature implements Feature {
+        FEATURE_1
+    }
+
+    private static class AmazonS3ClientMOCK extends AmazonS3Client {
+        Map<String, Map<String, S3Object>> repo = new HashMap<String, Map<String, S3Object>>();
```
hexmind: `Map<String, Map<String, S3Object>> data = new HashMap<>();`

s3/src/main/java/org/togglz/s3/S3StateRepository.java
```diff
@@ -0,0 +1,118 @@
...
+        this.client = builder.client;
+        this.bucketName = builder.bucketName;
+        this.keyPrefix = builder.keyPrefix;
+    }
+
+    @Override
+    public FeatureState getFeatureState(Feature feature) {
+        try (S3Object object = client.getObject(bucketName, keyPrefix + feature.name())) {
+            if (object != null) {
+                FeatureStateStorageWrapper wrapper = objectMapper.reader().forType(FeatureStateStorageWrapper.class).readValue(object.getObjectContent());
```
hexmind:
```
FeatureStateStorageWrapper wrapper = objectMapper.reader()
  .forType(FeatureStateStorageWrapper.class)
  .readValue(object.getObjectContent());
```

pom.xml
chkal: +1

s3/pom.xml
chkal: It is OK this way, because the version and the scope are managed in the parent pom.

s3/pom.xml
chkal: Same as for junit.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
chkal: I actually prefer the builder approach. This is how we do it for other repositories too. This way the builder can enforce that all required parameters are set and all values are valid.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
chkal: +1 for splitting into multiple lines. It is hard to read this way.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
chkal: I don't think that we will need a warning hear. This exception is basically thrown if no feature state has been saved for this feature yet, which is a common case. No other repository logs warnings in this case.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
chkal: +1 for both

s3/src/main/java/org/togglz/s3/S3StateRepository.java
chkal: -1
Other repositories use inner classes for the builder too. And actually it must not be package scoped, because otherwise the developer cannot access it.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
MarkRx: It is necessary to wrap the Exception in a RuntimeException since the interface does not contain a thrown Exception.

I don't like adding custom exceptions that add no value. The other collectors wrap a RuntimeException.

pom.xml
MarkRx: done

s3/src/main/java/org/togglz/s3/S3StateRepository.java
MarkRx: The other collectors use the Builder pattern. I did what they did.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
MarkRx: Correct. The exception is used as a not-found state. It's perfectly valid. I don't like this pattern but I didn't write the AWS library.

s3/src/main/java/org/togglz/s3/S3StateRepository.java
MarkRx: I just copied the pattern that the other implementations did

s3/src/main/java/org/togglz/s3/S3StateRepository.java
chkal: Sorry, my comment was a bit unclear. My point was that catching a `java.lang.Exception` is considered bad practice. Why did you add that catch block which basically catches EVERYTHING (include `NullPointerException` for example). Does the Amazon client throw more checked exception? In this case just catch these specific types. I don't see any value in catching "everything" here. So my votes goes for removing the block that catches `java.lang.Exception`.

MarkRx:
I see what you mean. Fixed.

chkal:
Sorry, I'm very busy at the moment. I'll come back to the PR as soon as possible...

chkal:
Sorry for the delay. I lay sick in bed and therefore wasn't able to respond earlier.

IMO the PR looks good now. There are some minor code formatting issues, but I'll fix this after merging.

Thank you very much for your contribution.

PR #197, opened 2017-01-07
Slack notifications

hexmind:
This togglz-slack module allows to notify your team about feature toggles changes. 
See [README.md](https://github.com/hexmind/togglz/blob/master/slack/README.md)


slack/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+  <dependencies>
+
+    <dependency>
+      <groupId>org.togglz</groupId>
+      <artifactId>togglz-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>17.0</version>
```
szpak: `17.0` is quite old (22 Apr 2014). Maybe it would be good to use newer version as you introduce it as a production dependency (currently Guava 17 is used only in a test scope for a core module).

slack/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+              <goal>testGenerateStubs</goal>
+              <goal>testCompile</goal>
+            </goals>
+          </execution>
+        </executions>
+        <dependencies>
+          <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>2.4.4</version>
```
szpak: `2.4.4` vs `2.4.7` declared above.

slack/src/main/java/org/togglz/slack/ChannelsProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+class ChannelsProvider {
+
+    private static final Log log = LogFactory.getLog(ChannelsProvider.class);
+
+    private final List<String> channels;
+
+    ChannelsProvider(List<String> channels) {
+        Preconditions.checkArgument(channels != null, "channels can be empty but not null");
+        this.channels = channels;
```
szpak: Defensive copy could be created

slack/src/main/java/org/togglz/slack/config/NotificationConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+        checkArgument(isNotBlank(slackHookUrl), "slackHookUrl is required");
+        checkArgument(slackHookUrl.startsWith("http"), "slackHookUrl isn't valid url: %s", slackHookUrl);
+        this.slackHookUrl = slackHookUrl;
+        this.channels = firstNonNull(channels, DEFAULT_CHANNELS);
+        this.togglzAdminConsoleUrl = firstNonNull(togglzAdminConsoleUrl, "");
+        this.appName = firstNonNull(appName, "");
+        this.messageFormat = firstNonNull(messageFormat, NotificationComposer.DEFAULT_MESSAGE_FORMAT);
+        this.appIcon = firstNonNull(appIcon, DEFAULT_APP_ICON);
+        this.stateIcons = firstNonNull(stateIcons, DEFAULT_STATE_ICONS);
+        this.changeVerbs = firstNonNull(changeVerbs, DEFAULT_CHANGE_VERBS);
```
szpak: Probably also a defensive copy

slack/src/main/java/org/togglz/slack/notification/Markdown.java
```diff
@@ -0,0 +1,49 @@
+package org.togglz.slack.notification;
+
+import com.google.common.base.Joiner;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public enum Markdown {
```
szpak: Nice idea

slack/src/main/java/org/togglz/slack/sender/HttpPostRequest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.togglz.core.logging.LogFactory;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+class HttpPostRequest {
```
szpak: Good to keep some classes package local.

slack/src/test/groovy/org/togglz/slack/config/NotificationConfigurationBuilderSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+class NotificationConfigurationBuilderSpec extends Specification {
+
+    static final String HOOK_URL = "https://hooks..."
+
+    def "should require withSlackHookUrl property"() {
+        when:
+            NotificationConfiguration.builder()
+                    .build()
+        then:
+            Exception e = thrown(IllegalArgumentException)
```
szpak: Can be simplified with:
```
IllegalArgumentException e = thrown()
```

slack/src/test/groovy/org/togglz/slack/config/NotificationConfigurationBuilderSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+            config.channels == ["channel"]
+            config.togglzAdminConsoleUrl == "console"
+            config.appName == "app"
+            config.appIcon == "icon"
+            config.getStateIcon(ENABLE_F1) == "+1"
+            config.getStateIcon(DISABLE_F1) == "-1"
+            config.getChangeVerb(ENABLE_F1) == "ON"
+            config.getChangeVerb(DISABLE_F1) == "OFF"
+            config.messageFormat == "format"
+            config.disabledAsyncSender
```
szpak: Maybe that?
```
with(config) {
    slackHookUrl == HOOK_URL
    ...
}
```

slack/src/test/groovy/org/togglz/slack/notification/MarkdownSpec.groovy
```diff
@@ -0,0 +1,47 @@
+package org.togglz.slack.notification
+
+import spock.lang.Specification
+import spock.lang.Unroll
+
+@Unroll
```
szpak: https://github.com/szpak/spock-global-unroll could be used to unroll all parameterized tests in the whole module

slack/template.mf
```diff
@@ -0,0 +1,12 @@
...
+Bundle-ManifestVersion: 2
+Bundle-SymbolicName: org.togglz.slack
+Bundle-Vendor: http://www.togglz.org/
+Bundle-Version: ${osgi.bundles.version}
+Bundle-Name: ${project.name}
+Version-Patterns:
+ default;pattern="[=.=.=, =.+1.0)"
+Import-Template:
+ org.togglz.core.*;version="${osgi.bundles.version:default}",
+ com.google.common.*;version="19.0",
```
szpak: `19.0` vs `17.0`. Probably 19 is a sensible default.

slack/src/main/java/org/togglz/slack/SlackStateRepository.java
```diff
@@ -0,0 +1,62 @@
...
+import org.togglz.slack.config.NotificationConfiguration;
+import org.togglz.slack.notification.Notification;
+import org.togglz.slack.notification.NotificationComposer;
+import org.togglz.slack.sender.AsyncNotifier;
+import org.togglz.slack.sender.NotificationSender;
+import org.togglz.slack.sender.Notifier;
+
+import java.util.List;
+
+public class SlackStateRepository implements StateRepository {
```
szpak: It's the main entry point to that feature. You could probably add a simple JavaDoc containing among others `@since` and `@author` tags.

szpak:
Very nice feature. Great that you buckled down to implement it!

slack/pom.xml
hexmind: It was intentional downgrade (to core module version).
@chkal,  is a good idea to upgrade Guava (core and slack modules) in next PR?

chkal: To be honest, I don't like having a dependency on Guava at all. Don't get me wrong, I love Guava. For my applications I use it very often. But I had major issues in the past where different 3rd party libraries depended on different older Guava versions which very totally incompatible. That's because the Guava guys remove deprecated functionality soon after the deprecation. And this can get you into Guava version hell. That's why I think that libraries (like Togglz) should not depend on Guava. What do you use Guava for?

slack/src/main/java/org/togglz/slack/SlackStateRepository.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public FeatureState getFeatureState(Feature ignored) {
+        return null;
+    }
+
+    @Override
+    public void setFeatureState(FeatureState state) {
+        List<Notification> notifications = composeNotifications(state);
+        log.info("send " + notifications);
```
chkal: This would get a bit nosy. I think `debug` level is better here.

slack/src/main/java/org/togglz/slack/sender/HttpPostRequest.java
chkal: +1

slack/src/main/java/org/togglz/slack/sender/HttpPostRequest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    HttpPostRequest(String requestUrl) {
+        this.requestUrl = requestUrl;
+    }
+
+    String send(byte[] body) {
+        HttpURLConnection connection = null;
+        try {
+            URL url = new URL(this.requestUrl);
+            connection = (HttpURLConnection) url.openConnection();
```
chkal: We should set connection and socket timeouts here. By default there are no timeouts, so in case of network problems this may hang forever, which is pretty long. ;-)

slack/src/main/java/org/togglz/slack/SlackWrapperStateRepository.java
```diff
@@ -0,0 +1,16 @@
+package org.togglz.slack;
+
+import org.togglz.core.repository.StateRepository;
+import org.togglz.core.repository.composite.CompositeStateRepository;
+
+/**
+ * @author Tomasz Skowroński
+ * @since 2.4.0
+ */
+public class SlackWrapperStateRepository extends CompositeStateRepository {
```
chkal: I think it is a bit confusing to have `SlackWrapperStateRepository` and `SlackStateRepository`. Especially because `SlackStateRepository` doesn't really implement something that can be considered a state repository. Why not merge these into a single `SlackStateRepository` class. Then you could just pass the "real" state repository into the `SlackStateRepository` constructor and the implementation would just delegate to the real repository before or after the notification is sent. So the `SlackStateRepository` would effectively just be a wrapper that adds notification behavior to an existing repository. That's IMO much simpler to use.

hexmind: Like this?
```
public class SlackStateRepository implements StateRepository {

    private final StateRepository wrappedRepository;
    private final SlackNotifications slackNotifications;

    public SlackStateRepository(StateRepository wrappedRepository, NotificationConfiguration configuration, UserProvider userProvider) {
        this.wrappedRepository = wrappedRepository;
        this.slackNotifications = new SlackNotifications(configuration, userProvider);
    }

    @Override
    public FeatureState getFeatureState(Feature feature) {
        return wrappedRepository.getFeatureState(feature);
    }

    @Override
    public void setFeatureState(FeatureState featureState) {
        wrappedRepository.setFeatureState(featureState);
        slackNotifications.notify(featureState);
    }
}
```
and package scope SlackNotifications (previously SlackStateRepository)

chkal: Yeah, that's what I was thinking about. IMO this makes much more sense, doesn't it? 

And it reduces the coupling to the StateRepository SPI. Which is a good thing because we may think about adding more explicit support for notifications to the core API at some point.

slack/pom.xml
hexmind: I use ByteStreams, CharStreams, FluentIterable and many common classes. I can try to replace them with java.util and org.togglz.core.util.
I can repackage some classes from Guava too (Apache License Version 2.0).

chkal: I know that using utility libraries like Guava is usually more convenient. But for library code it is very important do use as few dependencies as possible. Every dependency introduces potential version conflicts. And this is especially true for Guava, which removes deprecated methods/classes after a few releases. So if it is possible to adjust your code to rely on plain JDK classes, you should do so. If it is not possible for certain reasons, we could think about repackaging or just go with the Guava dependency ignoring potential problems in the future.

hexmind: I understand your motivation. Guava has been removed.

hexmind: @chkal is it OK now?

chkal:
Sorry, I'm very busy at the moment. I'll come back to the PR as soon as possible...

hexmind:
@chkal can you delegate last changes code review to szpak user?

chkal:
Hey @hexmind, sorry for the delay. I lay sick in bed and therefore wasn't able to respond earlier.

First of all: I like the new `SlackStateRepository` and how it wraps another repository. This looks much better and decouples the notification logic from the repository interface.

And I'm also happy to see that you removed Guava from the module dependencies. As I mentioned before reducing dependencies is IMO very important for a library like Togglz to reduce the risk of dependency conflicts for the users.

However, I was very surprised about the size of the pull request. There are 42 (!) changed files. After I had a closer look, I saw that you changed many unrelated files in the PR:

* You changed the signature of `Strings.join(Iterable, String)` to `Strings.join(String, Iterable)`. Therefore you updated many classes which were using the old signature. Why did you do so? I guess it is because you also added `Strings.join(Iterable, String...)` which only works if the varargs param is the last one. However, please don't do this signature change because it increases the size of the PR and makes it VERY hard to review. Please use something like `Strings.join( Arrays.asList(values), "," )` on the calling side instead.
* You have copied some classes from Guava to the source tree. Which is basically fine. But it looks like there are many methods which you are actually not using or which aren't really required. Like `MoreObjects.equal()` and `MoreObjects.hashcode()` which IMO are basically unnecessary because `java.lang.Objects` provides the same methods.
* You renamed an unrelated tests method in `SystemPropertyActivationStrategyTest`. Although I agree that the rename makes sense, please don't include it in this pull request. Especially because typically pull requests are squashed into a single commit when applied.

To sum up. Please keep the PR as small as possible and only include changes which are really required for the module to keep the review process as simple as possible.

hexmind:
Welcome back and sorry for non-PR changes. I applied your comments.

chkal:
Thanks a lot! I think we can merge this now. Great work! I'll give it a try myself as soon as I'm back in the office. We are using Rocket.Chat, which offers Slack compatible web hooks. ;)

Thanks a lot for your hard work.

hexmind:
When do you plan 2.4.0.RC1 ?

chkal:
I've to do some more cleanup.. Perhaps next week...

chkal:
BTW: The snapshots should be available in a few minutes...

PR #199, opened 2017-01-30
Activation strategies for DeltaSpike configuration properties and project stages

neocotic:
This PR adds support for feature activation based on DeltaSpike configuration properties (via `ConfigResolver`) and project stages (via `ProjectStage` and `ProjectStageProducer`). This was requested by @johnament based on my changes for #186 and #188. As part of this PR I have attempted to avoid duplication and encourage reusability by creating some abstract activation strategies to support tokenization of parameter values as well as property-based activation decisions.

I don't know a lot about DeltaSpike, nor have I ever used it before, but Togglz appears to depend on a very old version, 0.2-incubation. For the property resolution, I had wanted to use [ConfigResolver.getProjectStageAwarePropertyValue](https://deltaspike.apache.org/javadoc/1.2.1/org/apache/deltaspike/core/api/config/ConfigResolver.html#getProjectStageAwarePropertyValue-java.lang.String-) instead of just [ConfigResolve.getPropertyValue](https://deltaspike.apache.org/javadoc/1.2.1/org/apache/deltaspike/core/api/config/ConfigResolver.html#getPropertyValue-java.lang.String-) as I believe it could be more useful and flexible, however, this method was unavailable in the current version. It appears that togglz will break on anything newer than 0.2-incubation (even on 0.3-incubation) because the Identity Management (IDM) part of DeltaSpike was removed after that version, which leaves `DeltaSpikeUserProvider` in a state where it can't compile. Since DeltaSpike has moved far on since then, perhaps it would be best if we simply upgraded to at least v1.2.1 (the first version for which I could find API documentation generated) and removed `DeltaSpikeUserProvider`. As-is, the user provider is  completely broken for the majority of DeltaSpike users.


johnament:
@neocotic I agree with your synopsis.  I can look at what's different between the DS versions and let @chkal know what to upgrade to make the DS stuff work in the newer versions.  I always recommend people go to latest, we haven't changed much from a backwards compatibility standpoint.

Changes look great.  This will be an extremely useful feature.

johnament:
Oh I see now.  @chkal heh you're using basically the one thing we've ever pulled out of DS.  Maybe to avoid changing it, removing it from togglz (but I can't believe it would actually work) create a separate module for the configuration?

chkal:
@johnament: Yeah, your are right: Back then I used the security API which was later removed. I'm fine with updating to the latest DS version and simply removing my old security integration classes. I hope that nobody uses this old DS version today anyway. :laughing:

neocotic:
@chkal I've upgraded DeltaSpike to the latest version and removed the obsolete `DeltaSpikeUserProvider`. I've also changed the new activation strategy to use project-stage-aware property values instead as I believe that this will be more powerful and more in keeping with the spring implementation.

chkal:
Thank so much. I'll have a deeper look at the PR ASAP.

chkal:
Hey @neocotic! I'm soooo sorry. I totally forgot about this one. I'm really sorry about that.

I reviewed this today and it looks great! Outstanding work! Thanks a lot. I had to manually merge it due to some merge conflicts, but these were easy to fix! Thanks so much! This is really great!

neocotic:
@chkal, no problem at all. As I don't use DeltaSpike, I never actually missed it going in, however, I'm glad that you see the benefit in these changes and that I've implemented it correctly.

PR #200, opened 2017-01-31
Fixes FeatureVariations invoking BeforeClass and AfterClass methods multiple times

### LINKED ISSUE #185: togglz's support for junit @BeforeClass and @AfterClass is weird
Hi,

Here is a simple test we did, we are running with annotation `@RunWith(FeatureVariations.class)`: 

``` java
@Variations
public static VariationSet<EFeature> getPermutations() {
    return VariationSetBuilder.create(EFeature.class).vary(EFeature.NEW_FEATURE);
}
```

Our test is running in embedded glassfish as an EJB integration test. In our test, we simple do the following:

``` java
@BeforeClass
public static void setUpClass() {
    logger.log(Level.INFO, "Setup Class got called ===================> ");
}

@AfterClass
public static void tearDownClass() {
    logger.log(Level.INFO, "Teardown Class got called ===================> ");
}
```

This is what we got:

```
[#|2016-11-09T10:31:24.291-0800|INFO|undefined 0.0|BatchUploadingServicesITTest|_ThreadID=1;_ThreadName=main;_TimeMillis=1478716284291;_LevelValue=800;|Setup Class got called ===================> |#]
[#|2016-11-09T10:31:18.925-0800|INFO|undefined 0.0|BatchUploadingServicesITTest|_ThreadID=1;_ThreadName=main;_TimeMillis=1478716278925;_LevelValue=800;|Setup Class got called ===================> |#]
[#|2016-11-09T10:31:18.985-0800|INFO|undefined 0.0|BatchUploadingServicesITTest|_ThreadID=1;_ThreadName=main;_TimeMillis=1478716278985;_LevelValue=800;|Setup Class got called ===================> |#]


Starting EJBTest
[#|2016-11-09T10:31:24.227-0800|INFO|undefined 0.0|BatchUploadingServicesITTest|_ThreadID=1;_ThreadName=main;_TimeMillis=1478716284227;_LevelValue=800;|Teardown Class got called ===================> |#]
[#|2016-11-09T10:31:26.762-0800|INFO|undefined 0.0|BatchUploadingServicesITTest|_ThreadID=1;_ThreadName=main;_TimeMillis=1478716286762;_LevelValue=800;|Teardown Class got called ===================> |#]

[#|2016-11-09T10:31:26.762-0800|INFO|undefined 0.0|BatchUploadingServicesITTest|_ThreadID=1;_ThreadName=main;_TimeMillis=1478716286762;_LevelValue=800;|Teardown Class got called ===================> |#]
```

So, two questions over here:

1. Based on my understanding, the whole unit test should only be run twice since I only did the variation on one Feature. How does the logging indicating Before/After class methods get called three times?
2. Looks like all the BeforeClass methods got called together at beginning and then all the After class methods got executed. This could be an issue, since most of the basic configuration could be configured at BeforeClass method and then after all the tests are done, AfterClass method will clean up this config. Having BeforeClass methods running all together will only create duplicate Entries in the database or duplicateEntry exception if there is an uniqueKey.

Looking forward to any comments.

neocotic:
This PR fixes #185 by changing `FeatureVariations` so that it no longer extends from `Suite` as this was resulting in `BeforeClass` and `AfterClass` methods being invoked multiple times.


chkal:
Hey @neocotic! I just had a quick look at the patch. But you may be right. Perhaps using `Suite` isn't the best way to go. I'll have a deeper look at the patch ASAP. Thanks for your work on this!

chkal:
Hey @neocotic. I just had a deeper look at your implementation.

It indeed looks like this fixes the behavior described in #185. However, it looks like your solution also has a downside:

The test methods are executed once for each permutation of features, and that's exactly what we want, BUT it seems like JUnit reports only a single execution. With the old implementation Eclipse and IntelliJ  (and the XML file generated by Surefire) showed one test execution for each permutation. Which was great, because if one of the permutations fail, you can immediately see which one causes problems. With your patch applied there is just a single execution recorded. Not sure if this is acceptable to be honest. And I'm not sure what is causing that.

After I had a look at the old code, I started to remember why I was using `Suite` base class. I basically looks at JUnits `Parameterized` runner which is basically something very similar to what we want. See:

https://github.com/junit-team/junit4/wiki/Parameterized-tests

If I understand it correctly, `Parameterized` also executes `@BeforeClass` and `@AfterClass` multiple times. That's why I wonder if the behavior mentioned in #185 is really a bug or expected behavior.

neocotic:
@chkal that makes sense. My approach would take it more inline with JUnit's [Theories](https://github.com/junit-team/junit4/wiki/Theories).

Personally, I prefer my BeforeClass and AfterClass methods only being invoked once, but I can see the benefit of getting reports on specific failures. That said; my approach (and that of using the `Theories` runner) does tell you which test failed. Just like for any other test failure, you'll have to debug to investigate why it failed anyway.

I don't mind either way as I prefer to use the togglz rule over the runner as I prefer to be more explicit in my tests, so I've never really used the runner. I just wanted to try and fix #185.

chkal:
@neocotic I personally also prefer the `TogglzRule` for writing test. And I never used the `FeatureVariations` myself in real world projects.

To be honest, I'm not sure whether `BeforeTest`/`AfterTest` should run once or for every variation. I think there are pros and cons for both variants. I believe however, that each variation should be reported as a separate execution. Not sure to be honest whether or not this is possible with Theories.

It would be great to get some more feedback on what others think about #185. I'm also a bit unsure whether we can change the current behavior as it may break people's tests. Maybe we get some more feedback by @zhangchi.

darnoldt:
I think the reason for calling BeforeClass multiple times is, that the `VariationRunner` does not override the method `protected Statement classBlock(RunNotifier notifier)` from the base class as does the `BlockJUnit4ClassRunnerWithParameters`.
I think it should suffice to override this method as follows:
```
    @Override
    protected Statement classBlock(RunNotifier notifier) {
        return childrenInvoker(notifier);
    }
```

chkal:
This pull request is pretty old and doesn't merge without conflicts any more. So closing this for now.

PR #201, opened 2017-02-03
Update spring boot to 1.5

Epimetheus89:
Togglz uses org.springframework.boot.context.embedded.ServletRegistrationBean which had been deprecated in 1.4.x and was removed in 1.5.x.
Also the test package has been cleaned up in 1.4.x: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-1.4-Release-Notes#test-utilities-and-classes


marceloverdijk:
I think we need to upgrade indeed.
Only problem I see is that "old" users of 1.4 (which are probably most of the users) would not be able to upgrade to latest Togglz Boot Starter in that case.

chkal:
Just to make sure I understand this correctly. Spring Boot removes classes between minor releases? :open_mouth:

Looks like they aren't using semantic versioning.

marceloverdijk:
Actually Boot 1.4 has 2 `ServletRegistrationBean` classes. One in `org.springframework.boot.web.servlet` and the other one in `org.springframework.boot.context.embedded`. The latter one (used by Togglz) was already deprecated in 1.4 and removed in 1.5.
I merged this PR to use the correct one (`org.springframework.boot.web.servlet`) which should be correct for both 1.4 and 1.5 users.

Thanks @Epimetheus89  for the PR much appreciated!

marceloverdijk:
@chkal would be good have a new Togglz release soon, as currently Boot 1.5 users are blocked 😢

chkal:
Oh, OK. I'll see if I can push out a release next week.

Thanks for fixing this! :+1:

Epimetheus89:
Thanks for your fast reply and for merging!

PR #206, opened 2017-03-13
Use the more basic AmazonS3 interface instead of AmazonS3Client

dasniko:
Amazon has changed their API for S3 recently.
It's deprecated and discouraged to use `new AmazonS3Client(...)`, instead, you should use e.g. `AmazonS3ClientBuilder.defaultClient()` (or similar), which doesn't return an `AmazonS3Client` object, but an `AmazonS3`.
This commit changes just the usage from `AmazonS3Client` to `AmazonS3`, all other functionality remains untouched.

I really appreciate the S3StateRepository!! 👍


chkal:
Awesome! Thank you very much! :beers:

PR #220, opened 2017-05-04
Cassandra & Zookeeper: upgrade to latest clients and dependencies; code cleanup

cosminrentea:


cosminrentea:
Hello,

(unrelated to the current PR)
Is there any plan to add support for Redis as a StateRepository ?

chkal:
Thanks for the PR. Basically it looks good. Just one questions. There seem to be some major version updates. Are there any incompatibilities users must be aware of?

PS: Please don't open and quickly close so many pull requests. This creates a lot of noise with notifications. If you want to modify the PR content, just do a force push.

chkal:
Regarding Redis: No, there are no plans. Pull requests welcome!

cosminrentea:
Zookeeper update is minor, there are no problems with backwards-compatibility.

Regarding Cassandra: upgrading is done to the Astyanax latest published version (3.9.0).
That project was retired by Netflix in February 2016. It relies on Thrift, which will be removed from Cassandra 4.x, thus eventually switching to using the Datastax driver (CQL-based) would be a good idea.
However, the current upgrade from Astyanax 2.0.2 to 3.9.0 is still useful since it enlarges the compatibility "range" with Cassandra from [1.2, 2.0] to [1.2 , 2.x , 3.x]
AFAIK Currently most Cassandra installations are using versions 2.x and 3.x.
Cassandra-Unit, the test-dependency, was also upgraded, and the integration test is now done against an embedded Cassandra 2.x (before, it was 1.x) .

Thanks for the comment and reply, I will try to keep PRs more focused and generate less verbosity :)

chkal:
Thanks a lot for your reply. I just merged the changes.

Having smaller and more focused pull requests is typically better because there are much easier to review.

PR #224, opened 2017-05-11
Redis State Repository

cosminrentea:


redis/src/main/java/org/togglz/redis/RedisStateRepository.java
```diff
@@ -0,0 +1,171 @@
...
+    public static final String STRATEGY_FIELD = "strategy";
+    public static final String PARAMETERS_FIELD = "parameters";
+
+    protected final JedisPool jedisPool;
+    protected final JedisPoolConfig jedisPoolConfig;
+    protected final String hostname;
+    protected final String keyPrefix;
+    protected final MapSerializer mapSerializer;
+
+    public RedisStateRepository() {
```
chkal: Just wondering. Why do you provide this default constructor? Is this useful for users? Why not force them to configure the repository correctly using the builder?

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
```diff
@@ -0,0 +1,171 @@
...
+            final Map<String, String> map = jedis.hgetAll(keyPrefix + feature.name());
+            if (map == null || map.size() == 0) {
+                return null;
+            }
+            final FeatureState featureState = new FeatureState(feature);
+            featureState.setEnabled(Boolean.valueOf(map.get(ENABLED_FIELD)));
+            featureState.setStrategyId(map.get(STRATEGY_FIELD));
+            final String strategyParameters = map.get(PARAMETERS_FIELD);
+            if (strategyParameters != null) {
+                for (final Map.Entry<String, String> entry : mapSerializer.deserialize(strategyParameters).entrySet()) {
```
chkal: I'm just wondering if it makes sense to use the MapSerializer here. Wouldn't it be better to store all the parameters separately. Just like you already do it for "enabled" and "strategy-id"? Wouldn't this be more in the spirit of Redis?

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
cosminrentea: OK, removed the default constructor.

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
cosminrentea: Would you like to prefix all parameters with some string constant, like: "parameters-" , and store/retrieve them individually (so that no naming collision can occur) ?

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
chkal: Ok, great!

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
chkal: Yeah, I think this would make sense. That's how I do it in the Properties-based repository.

cosminrentea:
Changes done

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
```diff
@@ -0,0 +1,151 @@
...
+ * <p>
+ * The class provides a builder which can be used to configure the state repository instance
+ * (e.g. Redis hostname, Redis/Jedis configuration, serialization).
+ * </p>
+ *
+ * @author Cosmin Rentea
+ */
+public class RedisStateRepository implements StateRepository {
+
+    public static final String KEY_PREFIX = "togglz:";
```
chkal: A minor one. Could you either inline this or move the constant to the builder. I was actually very confused by this. At first my impression was that this is a fixed prefix for the keys, but actually it is just the default prefix used if the user didn't specify another one. Therefore I think it would be ok to either inline it or move it to the build and name it `DEFAULT_KEY_PREFIX`.

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Builder for a {@link RedisStateRepository}.
+     * <p>
+     * Can be used as follows:
+     * </p>
+     * <pre>
+     *      StateRepository stateRepository =
+     *         new RedisStateRepository.Builder().
+     *         hostname("host").
+     *         mapSerializer(DefaultMapSerializer.singleline()).
```
chkal: `mapSerializer()` has been removed, so you can remove this one too

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
```diff
@@ -0,0 +1,151 @@
...
+ */
+public class RedisStateRepository implements StateRepository {
+
+    public static final String KEY_PREFIX = "togglz:";
+    public static final String ENABLED_FIELD = "enabled";
+    public static final String STRATEGY_FIELD = "strategy";
+    public static final String PARAMETER_PREFIX = "parameter:";
+    public static final int PARAMETER_PREFIX_LENGTH = PARAMETER_PREFIX.length();
+
+    protected final JedisPool jedisPool;
```
chkal: I'm starting to ask myself, whether it wouldn't be much simpler if the user could just provide a `JedisPool` directly via the builder. Then we could remove `jedisPoolConfig` and `hostname` which makes the repository much simpler. For the user this could also be much simpler, because if he already has a `JedisPool` in his application, he could simply use this one as a shared pool between Togglz and his app. What do you think?

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
cosminrentea: Inlined it, since it was used only once in the code.

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
cosminrentea: Fixed

redis/src/main/java/org/togglz/redis/RedisStateRepository.java
cosminrentea: I added another method to the builder - taking a JedisPool as a parameter. 
If no JedisPool is explicitly set, then the JedisPoolConfig + hostname will be used.
Removed  also the "jedisPoolConfig" from the fields of the RedisStateRepository class, since only "jedisPool" is actually used after the initialization.
Please review the changes.
Thank you for the feedback!

chkal:
Awesome! Thanks a lot! This looks great! I'll merge it right away.

PR #231, opened 2017-07-04
Add CompositeFeatureProvider

dsyer:
A composite pattern is useful to support multiple options for
configuring features simultaneously.


dsyer:
That test failure looks like it is unrelated to my change.

chkal:
Looks great! Thanks a lot!

PR #232, opened 2017-07-05
Add support for configuring features via Spring Environment

dsyer:
This will be much more comfortable for Spring and Spring Boot users
since there are multiple ways to get properties into the
Environment (and using Spring Cloud you can update it at runtime
as well).

The only unfortunate side effect is that the existing configuration
at togglz.features.<FEATURE>=true|false had to be moved down a level,
i.e. togglz.features.<FEATURE>.enabled=true|false, so that the
other properties (strategy, groups, etc.) can all be bound to the same
object. Since this is 2.5.0 I hope that will be acceptable (people
using the existing configuration model will have to migrate).

I was going to add some documentation, but I couldn't find the source
code for that. Can someone point me at the right way to do it?

### DIFF
--- a/spring-boot-starter/pom.xml
+++ b/spring-boot-starter/pom.xml
@@ -13,7 +13,7 @@
   <description>Togglz - Spring Boot Starter</description>
 
   <properties>
-    <spring-boot.version>1.5.1.RELEASE</spring-boot.version>
+    <spring-boot.version>1.5.4.RELEASE</spring-boot.version>
     <thymeleaf-togglz.version>1.0.1.RELEASE</thymeleaf-togglz.version>
   </properties>
 
@@ -120,30 +120,19 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
+      <optional>true</optional>
     </dependency>
 
     <dependency>
-      <groupId>org.assertj</groupId>
-      <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-test</artifactId>
+      <optional>true</optional>
     </dependency>
 
     <dependency>
         <artifactId>spring-boot-starter-test</artifactId>
         <groupId>org.springframework.boot</groupId>
         <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-test</artifactId>
     </dependency>
 
   </dependencies>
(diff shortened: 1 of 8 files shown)


dsyer:
Example of changes needed to make existing apps work with the new configuration model: https://github.com/togglz/togglz-samples/pull/4. (The old enum-based thing still works, but this feels more idiomatically Spring Boot.)

marceloverdijk:
I think it looks good to merge this in.

@chkal it contains a non backward change regarding `togglz.featuresfewature1.=true|false` vs. `togglz.features.feature1.enabled=true|false` as @dsyer mentioned.

If we merge this we should also update https://www.togglz.org/documentation/spring-boot-starter.html.

chkal:
@dsyer Thanks a lot for creating this PR. I think the update of the model is acceptable if we add some documentation about this.

A perfect place to add some text about how to migrate would be the "Updating Notes" page:
https://github.com/togglz/togglz-site/blob/master/updating-notes.html.haml

It would be great if you could provide a pull request for that. Or provide some text about the change. That would really be helpful.

And as @marceloverdijk already mentioned the documentation may also be affected:
https://github.com/togglz/togglz-site/blob/master/documentation/spring-boot-starter.html.haml

bennetelli:
@dsyer I'm just diving into this Pull Request and was wondering why the TogglzTestExecutionListener class is necessary. Maybe you can give me some insights about it?

dsyer:
Your guess is as good as mine after all this time. It looks like it activates all the features that say they are on by default? So I guess they are not active by default in a test, despite having metadata to say that they are.

bennetelli:
@dsyer thanks for your fast response :-) 
Then I'm willing to delete this piece of code, because a user has got issues with it (see #525) and I'm also not a big fan of having spring-test stuff in production code/projects like togglz-spring-core.
I didn't notice any side effects when deleting it. Neither in the togglz code itself nor in the togglz-samples project. That's why I would like to go as described. In case you totally disagree.. that's your chance ;-)

dsyer:
I would assume that it was doing something useful. I wasn't really clear what the issue was in #525 TBH. Also I don't think there was a togglz-junit module at the time that I contributed that code, but now that there is, maybe that's where this code should be?

bennetelli:
At least it would be better to have it in that module. Unfortunately the togglz-junit project is framework independent and I don't want it to be dependent on springframework.

dsyer:
I think I know what it does BTW, and it seems like it's probably needed. If you enable or disable features in tests, then you want to reset the state to the default automatically in between test methods. I think that's what it is doing, so I'm surprised if you can remove it and it doesn't cause test failures in those samples. Did you try that?

I don't see any harm in it being here personally (the code will never be used in production, so I don't know where you got that idea from), but if you want to extract it to a separate togglz-spring-test module, I guess that would work as well.

bennetelli:
I got it from #525: 
"Why is this there? It is completely out of place and belongs in the test package togglz-junit5 or togglz-junit. Having this class here will fail running tests as application context will not be able to start up in our use case as it forces features state to be set while some (application) dependencies are missing:
    we store an audit history for each feature toggle change and require a user to be present (the user who initiated the change). As in the test execution listener no security context is setup it will fail with an NPE in our code."

dsyer:
I know, I read that too, but it didn't seem like there was any connection between that and you saying that someone was "using AbstractTestExecutionListener in production code to change feature states". The issue in #525 sounds like maybe that's a bug, not a reason to rip out a feature and discard it. We would need more detail though to be able to fix it.

PR #242, opened 2017-08-27
Fix 227 TogglzProperties does not have @Validated

sdavids:


chkal:
Thanks a lot! Looks good.

PR #247, opened 2017-09-21
Add A URL Parameter Activation Strategy

th3morg:
My team has found this activation strategy very useful for developing and testing incomplete features that are being released dark. The ability to quickly view the toggled feature in an on/off state has proved valuable during development and when sharing links with stakeholders. URL parameters are a simple mechanism and enable stakeholders to keep up with progress while end users still do not have the toggled feature. 

Since this project has been incredibly valuable to us, I wanted to contribute this activation strategy back to the core project. Thank you!

### DIFF
--- a/servlet/pom.xml
+++ b/servlet/pom.xml
@@ -40,11 +40,6 @@
       <artifactId>togglz-test-harness</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpclient</artifactId>
-      <version>4.5.2</version>
-    </dependency>
 
   </dependencies>
 </project>
--- a/servlet/pom.xml
+++ b/servlet/pom.xml
@@ -40,6 +40,11 @@
       <artifactId>togglz-test-harness</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.2</version>
+    </dependency>
 
   </dependencies>
 </project>
--- a/servlet/src/main/java/org/togglz/servlet/activation/UrlParameterActivationStrategy.java
+++ b/servlet/src/main/java/org/togglz/servlet/activation/UrlParameterActivationStrategy.java
@@ -20,7 +20,7 @@ import org.togglz.servlet.util.HttpServletRequestHolder;
 import javax.servlet.http.HttpServletRequest;
 
 public class UrlParameterActivationStrategy implements ActivationStrategy {
-	static final String PARAM_URL_PARAMS = "params";
+	public static final String PARAM_URL_PARAMS = "params";
 	private static final Pattern SPLIT_ON_COMMA = Pattern.compile("\\s*,\\s*");
 	private static final Pattern SPLIT_ON_EQUALS = Pattern.compile("\\s*=\\s*");
 
@@ -53,19 +53,8 @@ public class UrlParameterActivationStrategy implements ActivationStrategy {
 		String referer = request.getHeader("referer");
 		if (referer != null) {
 			try {
-				List<NameValuePair> refererParameters = new URIBuilder(referer).getQueryParams();
-				Map<String, List<String>> combinedRefererParameters = new HashMap<>();
-				for (NameValuePair pair : refererParameters) {
-					if (!requestParams.containsKey(pair.getName())) {
-						if (combinedRefererParameters.containsKey(pair.getName())) {
-							List<String> values = combinedRefererParameters.get(pair.getName());
-							values.add(pair.getValue());
-							combinedRefererParameters.put(pair.getName(), values);
-						} else {
-							combinedRefererParameters.put(pair.getName(), new ArrayList<>(Arrays.asList(pair.getValue())));
-						}
-					}
-				}
+				Map<String, List<String>> combinedRefererParameters = getRefererParameters(requestParams, referer);
+				
 				for (Map.Entry<String, List<String>> entry : combinedRefererParameters.entrySet()) {
 					List<String> val  = entry.getValue();
 					String[] values = new String[val.size()];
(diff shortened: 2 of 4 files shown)


th3morg:
It appears that there may be a configuration issue with travis-ci. The errors don't appear as if they would be caused by my changes.

error418:
Hi there! It seems, that openjdk7 builds fail on travisci (see travis-ci/travis-ci#7884). Using `openjdk7` as an alternative would be a solution to this problem.

The `oraclejdk8` build on the other hand fails with following error message:
```
[WARNING] Bundlor Warnings:
[WARNING]     The import of package org.apache.http does not specify a version
[WARNING]     The import of package org.apache.http.client.utils does not specify a version

[...]

[ERROR] Failed to execute goal org.eclipse.virgo.bundlor:org.eclipse.virgo.bundlor.maven:1.1.2.RELEASE:bundlor (bundlor) on project togglz-servlet: Bundlor returned warnings.  Please fix manifest template at '/home/travis/build/togglz/togglz/servlet/template.mf' and try again. -> [Help 1]
```

th3morg:
@error418 - I figured out how to properly update the template.mf, but the oraclejdk7 build still fails. jdk8 looks good

error418:
The jdk7 build should work again when PR #249 gets approved and merged. I think the decision of switching to openjdk7 for jdk7-builds should be made in a separate pull request by the maintainers.

error418:
@th3morg - #249 is now merged to master. Your build should be fine after pulling the latest changes into your branch

chkal:
Hey everyone. Sorry for joining the discussion so late. I'll review the code later today and let you know what I find... thanks..

servlet/pom.xml
```diff
@@ -40,6 +40,11 @@
       <artifactId>togglz-test-harness</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
chkal: Do be honest, I would like to avoid additional dependencies here. As you see `togglz-servlet` only depends on `togglz-core` and the Servlet API. I think you don't use much from this dependency and it should be easy to do it without these classes.

servlet/src/main/java/org/togglz/servlet/activation/UrlParameterActivationStrategy.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Objects;
+import java.util.regex.Pattern;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URIBuilder;
+import org.togglz.core.activation.Parameter;
+import org.togglz.core.util.Strings;
+import org.togglz.servlet.util.HttpServletRequestHolder;
+import javax.servlet.http.HttpServletRequest;
+
+public class UrlParameterActivationStrategy implements ActivationStrategy {
```
chkal: Could you perhaps apply the Toggz code format here?
See: https://github.com/togglz/togglz/blob/master/eclipse-code-format.xml

servlet/src/main/java/org/togglz/servlet/activation/UrlParameterActivationStrategy.java
```diff
@@ -0,0 +1,117 @@
...
+			try {
+				Map<String, List<String>> combinedRefererParameters = getRefererParameters(requestParams, referer);
+
+				for (Map.Entry<String, List<String>> entry : combinedRefererParameters.entrySet()) {
+					List<String> val  = entry.getValue();
+					String[] values = new String[val.size()];
+					requestParams.put(entry.getKey(), val.toArray(values));
+				}
+			} catch (URISyntaxException e) {
+				e.printStackTrace();
```
chkal: Either use a real logger to log this exception or simply ignore it. But add a comment explaining why you ignore it in the latter case.

servlet/src/main/java/org/togglz/servlet/activation/UrlParameterActivationStrategy.java
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+		}
+
+		return hasActivationUrlParameter(allowedParams, requestParams);
+	}
+
+	private Map<String, List<String>> getRefererParameters(Map<String, String[]> requestParams, String referer) throws URISyntaxException {
+		List<NameValuePair> refererParameters = new URIBuilder(referer).getQueryParams();
```
chkal: Looks like this is the only reason for you to include the additional dependency. I think we can avoid that. You can use `java.net.URI` get the the query string from the referer and then just split by `&` and then by `=` to get the values. You should also use `URLEncoder` to decode the percent-encoded values from the query string. Then we can remove the dependency.

servlet/src/test/java/org/togglz/servlet/activation/UrlParameterActivationStrategyTest.java
```diff
@@ -0,0 +1,203 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.mockito.Mockito.when;
+
+public class UrlParameterActivationStrategyTest {
```
chkal: Great test coverage! :+1:

servlet/src/main/java/org/togglz/servlet/activation/UrlParameterActivationStrategy.java
th3morg: The only reason I used apache commons is because of it's robustness. If you're comfortable with a more simple approach, I will do that.

servlet/src/test/java/org/togglz/servlet/activation/UrlParameterActivationStrategyTest.java
th3morg: Thank you!

th3morg:
I removed the Apache dependency, added a test for the in class parameter checking for an encoded character, imported the eclipse file settings and reformatted the files, as well as fixing the e.printStackTrace() call. I think that wraps up the requests. Let me know if there's anything else! Thank you!

chkal:
Thanks a lot! Unfortunately I'm quite busy for personal reasons at the moment. I'll have a look at the code ASAP.

chkal:
Thanks a lot for addressing my comments. I'll merge this now. But I'm thinking about renaming the strategy to something like "Query Parameter" instead of "URL Parameter". I'll let you know.

chkal:
I refactored the code a bit and renamed the strategy. Could you explain how the "Referer" header plays into all of this? How does this work in practice?

th3morg:
The referer header was a necessary implementation for asynchronous calls from the page after the initial page load. It was the best/easiest way for that problem to be solved from my perspective without having to use another strategy in tandem (which would kind of defeat the purpose).

chkal:
Ok, thanks a lot for the explanation.

PR #252, opened 2017-10-02
FIX #251 - include parent contexts when looking for a bean

### LINKED ISSUE #251: FeatureManager not found with hierahical contexts
With Togglz 2.4.1 and Spring Boot 1.5 sometimes there was an exception "Could not find the FeatureManager.". I realized it's happening because Togglz listens to the `ContextRefreshedEvent`. If you set `management.port` to the different one than `server.port`, Spring creates separate context just for management. If this context is refreshed after the application context it's overriding the binding.

The generic solution is to scan not only the base context but also to include parents.

jkubrynski:


chkal:
Thanks a lot! Good catch! The patch looks perfect!

PR #255, opened 2017-10-19
Adding support for features within system properties to the base class for properties.

### LINKED ISSUE #254: Align SystemPropertyActivationStrategy with AbstractPropertyDrivenActivationStrategy
Togglz had a pre-existing `SystemPropertyActivationStrategy` which handled reading system properties.  A new base class `AbstractPropertyDrivenActivationStrategy` was added for property based toggles, however this class wasn't changed in the System Property one.

There's two key differences (one positive on each side):

- `SystemPropertyActivationStrategy` allows you to define the value to compare to.  `AbstractPropertyDrivenActivationStrategy` is hard coded to `true`.  It would be good if both supported custom property value comparisons.
- `AbstractPropertyDrivenActivationStrategy` supports automatic property naming, e.g. `togglz.FEATURE_NAME` when the property name isn't specified.  It would be good if `SystemPropertyActivationStrategy` supported this as well.

It seems to me the logical path would be to move value customization into the parent class and make system property extend from that class, this way they get the same feature set.

johnament:
- Base class now supports empty property checking
- All properties now handle custom value checking

This closes #254

### DIFF
--- a/core/src/main/java/org/togglz/core/activation/AbstractPropertyDrivenActivationStrategy.java
+++ b/core/src/main/java/org/togglz/core/activation/AbstractPropertyDrivenActivationStrategy.java
@@ -39,8 +39,10 @@ import org.togglz.core.util.Strings;
  */
 public abstract class AbstractPropertyDrivenActivationStrategy implements ActivationStrategy {
 
-    public static final String DEFAULT_PROPERTY_PREFIX = "togglz.";
+    private static final String DEFAULT_PROPERTY_PREFIX = "togglz.";
+    private static final String DEFAULT_STATE_VALUE = "true";
     public static final String PARAM_NAME = "name";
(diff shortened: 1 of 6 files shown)


chkal:
Hey @johnament! Thanks a lot! Basically this PR looks great! But it looks like your changes broke one of the Spring tests. See:

https://travis-ci.org/togglz/togglz/builds/290109138

johnament:
There's two things I'm not sure I like:

- Test duplication.  Should I move the parameter checking to an abstract class to verify it consistently?
- If there should be more tests around when the property is or isn't set in the SystemProps one.

chkal:
Thanks John! That looks great!

I think the tests are fine. I agree that it creates some test duplication, but I don't think that this is a real issue. And of cause more tests would be great, but this could be added later on..

Merging...

PR #257, opened 2017-11-05
Clean up tooglz-core code

Pryanic:


Pryanic:
@chkal  Please look at it

chkal:
Sorry, but I won't merge this one. The diff is too massive. And there are formatting changes I don't like (like the mepty `<p>` elements in the javadocs). Before cleaning up code formatting we need to create a consistent code format profile for both and Eclipse and IntelliJ. Then we can proceed with cleanup.

However, I agree that using the diamond operator instead of explicit generics is a great idea as we dropped JDK6 support a while ago. So a pull request for that change is highly welcome.

PR #258, opened 2017-11-12
Add JUnit 5 Extensions

### LINKED ISSUE #245: JUnit 5 support
New way to set featuretoggles during testing is needed since JUnit 5 does not support Rules anymore,
http://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4

rweisleder:
This PR adds support for JUnit 5. Have a look at the test cases how to use the extensions.

Basically `@AllEnabled` and `@AllDisabled` are the same as `TogglzRule.allEnabled` and `TogglzRule.allDisabled`. To enable a single feature among all disabled features, the `TestFeatureManager` is available as [test parameter](http://junit.org/junit5/docs/current/user-guide/#writing-tests-dependency-injection).

`@VaryFeatures` is basically the same as `@RunWith(FeatureVariations.class)`.
I reused the `VariationSet` and `VariationSetBuilder` from `togglz-junit`. But I would be happier to move these classes to togglz-testing and omit the dependency to `togglz-junit`.

JUnit 5 requires Java 8 at runtime, so the compile source and target for the new maven module is 1.8.
The CI build will fail with JDK 7!

If merged this resolves #245

### DIFF
--- a/junit/src/main/java/org/togglz/junit/vary/VariationSetBuilder.java
+++ b/junit/src/main/java/org/togglz/junit/vary/VariationSetBuilder.java
@@ -1,148 +1,55 @@
 package org.togglz.junit.vary;
 
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.Set;
-
 import org.togglz.core.Feature;
-import org.togglz.core.util.Validate;
 
 /**
- * Default implementation of {@link VariationSet} that allows to build the set dynamically.
- * 
- * @author Christian Kaltepoth
- * 
+ * @see org.togglz.testing.vary.VariationSetBuilder
  * @param <F> The feature class
+ *
+ * @deprecated use {@link org.togglz.testing.vary.VariationSetBuilder} instead
  */
-public class VariationSetBuilder<F extends Feature> implements VariationSet<F>
-{
+@Deprecated
+public class VariationSetBuilder<F extends Feature> extends org.togglz.testing.vary.VariationSetBuilder<F>
+    implements VariationSet<F> {
 
-    public static <F extends Feature> VariationSetBuilder<F> create(Class<F> featureClass)
-    {
+    /**
+     * @deprecated use {@link org.togglz.testing.vary.VariationSetBuilder#create(Class)} instead
+     */
+    public static <F extends Feature> VariationSetBuilder<F> create(Class<F> featureClass) {
         return new VariationSetBuilder<F>(featureClass);
     }
 
-    private final Class<F> featureClass;
-
-    private final Set<F> featuresToVary = new HashSet<F>();
-    private final Set<F> featuresToEnable = new HashSet<F>();
-    private final Set<F> featuresToDisable = new HashSet<F>();
-
-    private VariationSetBuilder(Class<F> featureEnum)
-    {
-        Validate.notNull(featureEnum, "The featureEnum argument is required");
-        Validate.isTrue(featureEnum.isEnum(), "This class only works with feature enums");
-        this.featureClass = featureEnum;
+    private VariationSetBuilder(Class<F> featureEnum) {
+        super(featureEnum);
     }
 
-    /**
-     * Vary this feature in the variation set.
-     */
-    public VariationSetBuilder<F> vary(F f)
-    {
-        featuresToVary.add(f);
-        featuresToEnable.remove(f);
-        featuresToDisable.remove(f);
-        return this;
-    }
-
-    /**
-     * Enable this feature in the variation set.
-     */
-    public VariationSetBuilder<F> enable(F f)
-    {
-        featuresToVary.remove(f);
-        featuresToEnable.add(f);
-        featuresToDisable.remove(f);
-        return this;
-    }
-
-    /**
-     * Disable this feature in the variation set.
-     */
-    public VariationSetBuilder<F> disable(F f)
-    {
-        featuresToVary.remove(f);
-        featuresToEnable.remove(f);
-        featuresToDisable.add(f);
-        return this;
-    }
-
-    /**
-     * Enable all features in the variation set.
-     */
-    public VariationSetBuilder<F> enableAll() {
-        for (F f : featureClass.getEnumConstants()) {
-            enable(f);
-        }
-        return this;
-    }
-
-    /**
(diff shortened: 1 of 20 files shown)


chkal:
Thanks a lot for the pull request. I'll have a deeper look and review it ASAP. But it may take some days.

chkal:
Sorry for the delay. I still didn't find to have a deeper look at this. Sorry about that! :cry:

rweisleder:
@chkal any news on this?

chkal:
Hey @rweisleder. I'm really really sorry for the delay. I didn't find any time yet to have a deeper look at the pull request. I'm very sorry. My wife is currently hospitalized and I've to take care of the kids beside my regular job. That's why I don't find much time for Togglz at the moment. Anyway, I really appreciate the work you've done and I'm trying my best to review it as soon as possible.

junit-extension/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.togglz</groupId>
+    <artifactId>togglz-project</artifactId>
+    <version>2.6.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>togglz-junit-extension</artifactId>
```
chkal: I would prefer the artifact name `togglz-junit5`. Having `togglz-junit` and `togglz-junit-extension` would be confusing. But I think if we have `togglz-junit` and `togglz-junit5`, people will understand the difference. Please also update the name of the directory to `junit5`.

junit-extension/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+          <useDefaultManifestFile>true</useDefaultManifestFile>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.eclipse.virgo.bundlor</groupId>
+        <artifactId>org.eclipse.virgo.bundlor.maven</artifactId>
+        <executions>
+          <execution>
+            <id>bundlor</id>
+            <phase>none</phase>
```
chkal: I think it is ok to disable it. Especially as it is also disabled in the `junit` module. Not sure what the OSGi guys think about that!?!

junit/src/main/java/org/togglz/junit/vary/VariationSetBuilder.java
```diff
@@ -140,7 +140,8 @@ else if (featuresToVary.contains(feature)) {
         return copy;
     }
 
-    public Class<? extends Feature> getFeatureClass()
+    @Override
+    public Class<F> getFeatureClass()
```
chkal: Good catch! :+1:

pom.xml
```diff
@@ -184,6 +184,7 @@
     <module>mongodb</module>
     <module>testing</module>
     <module>junit</module>
+    <module>junit-extension</module>
```
chkal: You could move this into the [java8-features](https://github.com/togglz/togglz/blob/a0d1817cfd5adf746115b3e508abe914654b90f2/pom.xml#L300-L308) profile. This is the way we are currently handling modules which require JDK8.

rweisleder:
Thanks, @chkal, for the review! I implemented your change requests. If you have any other comments just let me know, especially how to handle the mentioned dependency between `togglz-junit` and `togglz-junit5`.
And the build is green now. 😄

junit5/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+
+  <dependencies>
+
+    <dependency>
+      <groupId>org.togglz</groupId>
+      <artifactId>togglz-testing</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.togglz</groupId>
+      <artifactId>togglz-junit</artifactId>
```
chkal: I wonder if this is problematic. Actually having both `togglz-junit` and `togglz-junit5` on the classpath seems weird. Especially because you will get all sorts of `togglz-junit` classes on your classpath although using them would fail because junit4 is missing. This may be confusing for users.

Thinking about this more: Wouldn't it make sense to move the `VariationSet` and `VariationSetBuilder` classes to `togglz-testing`? This way `togglz-junit5` could simply depend on this common testing module just like `togglz-junit`. AFAIK those classes actually doesn't depend on any JUnit API and therefore actually don't belong into the `togglz-junit` module. What do you think?

rweisleder: I agree, these classes should go to `togglz-testing`, they don't belong to JUnit. But into which package, `org.togglz.testing` or `org.togglz.testing.vary`?
Anyway, all projects using these classes need to adjust the imports in their (test) code. I think this should be acceptable.

chkal: I would prefer the package `org.togglz.testing.vary`.

But yes, your are correct. Moving these classes would break existing user tests which depend on these classes. I don't think that we can do that in a minor release. But perhaps we could use some tricks to preserve compatibility:

The `VariationSet` from `togglz-junit` could just extend the `VariationSet` from `togglz-testing`. Something like that:

```java
public interface VariationSet<F extends Feature> extends org.togglz.testing.vary.VariationSet<F>{}
```

And `VariationSetBuilder` from `togglz-junit` could just be an empty class with just a single static `create()` method which returns an instance of `VariationSetBuilder` from `togglz-testing`. The `create` method is the only way to create instances of this class, so this may work.

Then we could simply deprecate both classes from `togglz-junit` and therefore won't break user tests. Do you think this could work? That's just my thought without having a deeper look at the classes.

rweisleder: Sounds like a good idea, I will implement it this way.

chkal:
Hey @rweisleder. Awesome! Thank you so much. Sorry, I missed that you already pushed the requested changes. GitHub doesn't send any email notifies if you just push changes to the branch without adding a comment to the pull request.

This looks perfect. Moving the `VariationSetBuilder` totally makes sense and the structure is much cleaner now. So thanks again and sorry for the delay. I'll merge this immediately.

Issue #213, opened 2017-04-06
Consider using WebJars for JS/CSS resources


php-coder:
Please, consider using WebJars for static assets. The benefits:
1) applications that already use WebJars won't have 2 copies of the same resources
2) it will become possible to use the same version of Bootstrap/jQuery across the entire application
3) you won't have to hold sources of other projects in the Togglz repository

See for details: http://www.webjars.org/documentation

CC @dasniko

chkal:
Sounds like a good idea. Any other opinions?

dasniko:
yes, we can give it a try.

dasniko:
Just stumbled over this again...
I just had two thoughts on it, and wondering, if it's really a good fit:
* `togglz-console` ist just a JAR file, so, no `webjars` libs would be packed with it. When using it, users must include `webjars` deps themselves, might lead to errors/problems. No out-of-the-box usage would work
* Paths to `webjars` resources would include the version number, e.g. `webjars/bootstrap/3.3.7/css/bootstrap.min.css` - if users already use webjars, but in another version, it won't work. Users would have to use "our" webjars version, which is kind of force I don't like.

Please correct me, if I'm wrong at any of these points and/or how to solve it.

php-coder:
@dasniko Thanks! They're really good points and a food for thoughts.

The first point could be solved by having 2 jars -- one, like we have now, and another `togglz-console-nodeps` that requires user to include some deps on his own).

The second one is more complicated, but looks like https://github.com/webjars/webjars-locator-core could help with it.

dasniko:
@php-coder thanks for pointing me to the `webjars-locator-core`, didn't know this.
I will have a look into it and do some tests. Not yet convinced, especially with having two jars. But, let's see.
Hopefully it won't take another 7 months again ;)

php-coder:
@dasniko Thanks!

If I have to choose between waiting another 7 months and having this feature not implemented, I'd definitely pick the first :)

chkal:
Hey guys. Thanks a lot for continuing the discussion about this. I just posted a few comments on #262.

I know that 7 months are quite some time. But as you may have noticed I don't find much time to work actively on Togglz any more. Therefore any help like such pull requests is highly welcome!

chkal:
Hey all. Unfortunately we had to drop #262 as it was causing some issues. See the comments on #262 for details.

IMHO there are quite some issues with introducing webjars:

[...]

PR #262, opened 2017-12-04
Use webjars for static js/css assets

dasniko:
This is a first draft, like discussed in #213. Please test it and give feedback!


chkal:
Looks good to me. But I just checked the diff and didn't test it locally yet. WebJarAssetLocator is nice. Didn't know about this one.

The only downside I see with this PR is that we are basically dropping Servlet < 3.0 support. Not sure if it makes sense to still support Servlet 2.5, it just came to my mind while reading the diff. Any thoughts about that?

dasniko:
Yes, you're right. Didn't consider this. Do you have an idea of how widely servlet 2.5 is still used?

Another downside came up to me this moring: we're relying on e.g. Bootstrap 3.x, but 4.x is just to release and it changes the structure/classes (don't know how much, didn't look into it yet). Having different Bootstrap versions for 3.x in the project is ok, just anyone wins and it'll work. But if a project pulls in a 4.x resource, our layout might crash!

chkal:
> Yes, you're right. Didn't consider this. Do you have an idea of how widely servlet 2.5 is still used?

Good question. Not sure about that. One of the apps I'm maintaining is using JBoss 6.1.0 which for some reason doesn't support `META-INF/resources` correctly. I had to write a Servlet manually which does what the spec actually requires. However, I would be fine with dropping support for Servlet 2.5. Libraries dropping support for old spec versions are a nice way to force users to update their app servers. :-)

> Another downside came up to me this moring: we're relying on e.g. Bootstrap 3.x, but 4.x is just to release and it changes the structure/classes (don't know how much, didn't look into it yet). Having different Bootstrap versions for 3.x in the project is ok, just anyone wins and it'll work. But if a project pulls in a 4.x resource, our layout might crash!

Agreed! One way to work around this would be to shade the webjar, but I'm not sure if this is a good thing to do.

dasniko:
It will definitely break our layout, I just did a quick test. :-(

Shading would make no difference from the current solution, to have the distinct resources packed and delivered with the console. Shading would also lead to duplicate resources if someone uses another Bootstrap library.

So, I'm voting for dropping this PR and closing #213 for now, as it might lead to too much trouble/errors.

chkal:
I agree. Thanks anyway for prototyping this. At least we all know about `WebJarAssetLocator` now. ;-)

Issue #248, opened 2017-09-21
Don't use style attribute in web console to fix errors from 


php-coder:
I'm trying to integrate Content Security Policy (see https://content-security-policy.com) into my project and I found that browser complains about using `style` attributes in Togglz web console:

- `display: none;` (Bootstrap provide CSS class for that: https://stackoverflow.com/questions/18568736/how-to-hide-element-using-twitter-bootstrap-3-and-show-it-using-jquery)
- `width: 100%; text-align: center;`

Could you replace them by CSS classes, please?

chkal:
I'm not very familiar with Content Security Policy, but why are inline styles forbidden?

php-coder:
It is forbidden to reduce a possibility of injecting something malicious into it (similar to XSS attacks), like here: https://html5sec.org/#css This is from security perspective.

There are also others reasons to not use it: https://stackoverflow.com/questions/2612483/whats-so-bad-about-in-line-css

chkal:
To be honest, I don't agree with these arguments. The console contains static inline styles, so there is no danger of XSS and also using inline styles is IMO fine if you specify some style attributes for a single component and don't plan to reuse them.

However, pull requests for fixing these issues are highly welcome. So if anyone wants to contribute some code, this is a perfect opportunity. :smile:

dasniko:
In current version (2.5), I only can find `style="display: none;`.
The `width` and `text-align` seem to be gone with some refactoring.
Can you confirm this, @php-coder ?

php-coder:
@dasniko I can't confirm because I'm still on the old version. I'll try to update to a new version on the weekend and let you know.

dasniko:
I think, there was only one occurence of `width` and `text-align` and I eliminated it with this commit: https://github.com/togglz/togglz/commit/4fdc7438756f471b468d0241efbb19f5065ed3a2#diff-2a76b205759147bee0654ac1560639e4L26

So, this should be fine as I removed also the `display` occurence with the linked PR.
If there are no objections, I'll merge this soon.

chkal:
Thanks a lot! #248 looks good and should fix this problem.

php-coder:
> In current version (2.5), I only can find style="display: none;.

@dasniko I tested it and can confirm that there is only one left in 2.5 version.

dasniko:
thanks

PR #263, opened 2017-12-05
remove leftover style attribute in html code (#248)

dasniko:
see #248


chkal:
Looks great! Thanks a lot for addressing this one @dasniko.

PR #266, opened 2017-12-15
Logging state repository

margue:
Recently we had the requirement to log a short message, every time a feature is toggled, stating the feature's name and the new state. We implemented a very simple LoggingStateRepository for it.

Don't know, if this feature is worth being part of Togglz. If yes, here is a PR.


chkal:
Thanks a lot for providing the pull request.

I'm also not sure if the code will be useful for others, but it looks good so I'll merge it. Thanks a lot.

taaha47:
trying to log before delegating the action to the wrapped StateRepository means that a message will be logged before the state of the feature has been effectively changed, in case of an exception in the wrapped StateRepository for example. 
I believe the delegation operation must be the first line inside the setFeatureState method.

Issue #269, opened 2018-03-26
FeatureProvider implementations should not allow its clients to change the default FeatureState in their internal metaDataCache


GrahamLea:
If the first time a FeatureState is retrieved from the FeatureManager.getFeatureState() that FeatureState object is used to change the state (FeatureState.setEnabled()), then that call changes the value of the DefaultFeatureState in the internal cache of metadata in EnumBasedFeatureProvider.metaDataCache / PropertyFeatureProvider.metadata.

Implementations of FeatureMetaData should not allow their internal state (defaultFeatureState) to be modified.

chkal:
Thanks! I just merged your PR!

PR #270, opened 2018-03-26
#269: Tests and bug fix to prevent FeatureProvider clients from inadvertently changing the default FeatureState

GrahamLea:


chkal:
Good catch! Thanks a lot!

PR #272, opened 2018-04-20
Calling getFeatureState twice

luolong:
As part of attempting to debug an issue with composite feature states, I found that `CompositFeatureState.getFeatureState(feature)` calls `getFeatureState()` of the child repository twice.

This seems like a benign thing, but it may result in doubling of the read load on the child feature store.


core/src/main/java/org/togglz/core/repository/composite/CompositeStateRepository.java
```diff
@@ -74,7 +74,7 @@ public FeatureState getFeatureState(Feature feature) {
         for (StateRepository repository : iterationOrder.getSelected(repositories)) {
             FeatureState featureState = repository.getFeatureState(feature);
             if (featureState != null) {
-                return repository.getFeatureState(feature);
+                return featureState
```
chkal: There is a `;` missing...

chkal:
Thanks a lot!

Issue #279, opened 2018-06-29
Spring boot dedicated management context causes Tooglz warning during startup


jpmsilva:
With Togglz 2.5.0.Final and Spring Boot, if we setup the actuator endpoints in a dedicated context (by using the property `management.server.port` to specify a separate port), Spring will emit two `ContextRefreshedEvent` events, one for the main context and another for the management context.
This results in the following warning by Tooglz in the log:

`2018-06-29 13:52:41.414  WARN 29069 --- [           main] licationContextBinderApplicationListener : ApplicationContext already bound to current context class loader, releasing it first`

I have placed a breakpoint in `TogglzApplicationContextBinderApplicationListener.onApplicationEvent` and run several tests, and it appears that the management event is always received first.
This means that after the application starts up, the `ContextClassLoaderApplicationContextHolder` ends up containing the correct (main) application context.
However, I cannot be certain that the order of these events is always the observed one, and I don't think Spring Boot makes no such guarantees, which means that even if the order is correct now, it may not be in some future version.

I think that Togglz should not rely on the order of these events, and I propose that `ContextRefreshedEvent` events originating from the management context should instead be ignored.

If you agree with this idea, I can work on a PR to implement it.

Thanks and best regards

chkal:
Thanks for reporting this!

Could you perhaps give `2.6.0-SNAPSHOT` a try? I was actually planing to release `2.6.0.Final` this weekend. And @ruifigueira spent quite some time reworking the Spring Boot support. Especially for the new Spring Boot 2.0 support.

@ruifigueira Maybe we should fix this before pushing out 2.6.0?

jpmsilva:
Hi @chkal 

Just tested with a fresh build of 2.6.0-SNAPSHOT from the master branch, and I have the exact same result:

`2018-06-29 17:13:42.292  WARN 13826 --- [           main] licationContextBinderApplicationListener : ApplicationContext already bound to current context class loader, releasing it first`

I don't know if @ruifigueira will be able to reply soon (we actually work together). I'll ask him if he has some spare time to look at this.

jpmsilva:
I took a quick look at the code and I have no good news :disappointed: 

The problem I have is that identifying which context a ContextRefreshedEvent belongs to depends on casting the source of the event to a `org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext`, and comparing it's `serverNamespace` to the value `management`.
[...]

PR #280, opened 2018-06-29
Tentative fix for #279

jpmsilva:
This provides a tentative fix for #279 by introducing the concept of `ContextRefreshedEventFilter`.
Each Spring Boot starter (1.x and 2.0) provide their specialized versions of the filter, to ensure that only the main application context is captured by `ContextClassLoaderApplicationContextHolder`.
In the absence of such a filter, it falls back to the current behavior, which is to capture any application context carried by `ContextRefreshedEvent` events.

We should probably wait for @ruifigueira to get back and give his opinion.


jpmsilva:
I found some hardcoded references to Spring 3.0.7-RELEASE in test classes (SpringBasicOperationTest.java, SpringEarlyFeatureUsageTest.java, ManagedFeatureManagerTest.java and FeatureProxyTest.java).
The ASM library (shaded) bundled with this version of Spring cannot process Java 8 bytecode containing lambdas, and as such is causing these tests to fail.
I will check how to align the version of Spring used to package test.war with the declared version at the top level pom.xml, so that this issue will not recur.

chkal:
Oh yeah, these hardcoded version numbers should be eliminated if possible. IIRC Shrinkwrap can resolve artifacts versions from the pom, but I don't remember the details any more.

jpmsilva:
Hi @chkal 
I replaced the version reference to a call to `SpringVersion.getVersion(), which is a facility given by Spring itself to return the current running version, as specified by `Implementation-Version` from the Spring's JAR manifest. This has fixed the issue and allows the tests to complete.
However, #285 has now become a pre-requirement for this PR.
Waiting for @ruifigueira to give me his input on this.

chkal:
@jpmsilva Thanks a lot! I just merged #285. Maybe we should also wait for #283 and then rebase this pull request? Just to make sure all the tests still pass.

jpmsilva:
@chkal I will defer to @ruifigueira because he mentioned that same reasoning, so I suppose we will wait for #283 to be merged.
I already did the exercise of rebasing this PR on top of that one, and all tests were passing, which I think will speed up the process after.

chkal:
We have merged #283. Could you rebase against `master`?

@ruifigueira What do you about the PR?

ruifigueira:
Hmm, after better looking into `TogglzApplicationContextBinderApplicationListener`, that would not be as easy as I thought, and I guess that's why you included the filter interface in the first place, so I'm ok with the solution. I think that once it is rebased, we can merge it.

jpmsilva:
Hi @chkal and @ruifigueira 
After rebasing, all tests are passing.
If all is well, you could consider cutting a new version for general use :smile: 
Thanks

chkal:
Thanks a lot! As @ruifigueira is +1 for merging this, I'll do so now!

Issue #282, opened 2018-07-01
Spring test issue after upgrade to 2.6.0


amir-khassaia:
Hi, Upgrading from 2.5.0.Final to 2.6.0.Final shows up a runtime error with spring tests:
Running on Java 8, Spring 4.3.x

```
java.lang.IncompatibleClassChangeError: Found interface org.springframework.test.context.TestContext, but class was expected

	at org.togglz.spring.test.TogglzTestExecutionListener.beforeTestMethod(TogglzTestExecutionListener.java:30)
	at org.springframework.test.context.TestContextManager.beforeTestMethod(TestContextManager.java:269)
	at org.springframework.test.context.junit4.statements.RunBeforeTestMethodCallbacks.evaluate(RunBeforeTestMethodCallbacks.java:74)
	at org.springframework.test.context.junit4.statements.RunAfterTestMethodCallbacks.evaluate(RunAfterTestMethodCallbacks.java:86)
	at org.springframework.test.context.junit4.statements.SpringRepeat.evaluate(SpringRepeat.java:84)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:252)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:94)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.springframework.test.context.junit4.statements.RunBeforeTestClassCallbacks.evaluate(RunBeforeTestClassCallbacks.java:61)
	at org.springframework.test.context.junit4.statements.RunAfterTestClassCallbacks.evaluate(RunAfterTestClassCallbacks.java:70)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.run(SpringJUnit4ClassRunner.java:191)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
[...]

PR #283, opened 2018-07-02
Update minimum Spring version to 4.x

ruifigueira:
See:

* https://github.com/togglz/togglz/issues/282
* https://github.com/springtestdbunit/spring-test-dbunit/issues/46

Fixes: #282

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
     <!-- Set version of OSGi bundles without 'SNAPSHOT', etc. -->
     <osgi.bundles.version>2.6.0</osgi.bundles.version>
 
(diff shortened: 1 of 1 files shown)


chkal:
Does Spring use semantic versioning? I'm just wondering if we should update to the latest minor release of 4.x or if we should go with the latest 4.0.x release. Just to make sure that we don't depend on new features introduced in 4.3.x...

Any thoughts? But I would also be fine with updating to this version. Just want to share my thoughts. ;-)

chkal:
Ok, great! I agree. Let's merge this one.

PR #285, opened 2018-07-02
Removed hardcoded reference to Spring 3.0 from some tests.

jpmsilva:
This PR removes the hardcoded Spring 3.0 reference by calling `org.springframework.core.SpringVersion.getVersion()'.
Confirmed that this fixes the issue by running #280 with this patch.


chkal:
Looks good to me. I would actually prefer to use the metadata from the pom which is supported by recent Shrinkwrap version, but we can address this later when we find some time to update Arquillian and friends.

jpmsilva:
Hi @chkal 
Added PR #287 with an enhancement to resolve dependency versions resorting to the project's own `pom.xml` file.
I hope this is what you were thinking about.

PR #287, opened 2018-07-02
Allow resolving dependencies by using the project's own pom.xml

jpmsilva:
As commented in #285 this provides a more elegant solution to resolve dependency version from the project's own `pom.xml`


chkal:
Awesome! That's exactly what I had in mind. Great to see that older Shrinkwrap versions also support reading artifact metadata from pom files.

One idea: I think the behavior triggered by calling `filesystemRelativePomPath` should be the default. This way users of the `Deployments` class won't need to do anything and will get the metadata from the current pom for free. This would simplify the `Deployments` class and also the code for the integration tests. I see no reason why we should not want this metadata. So perhaps we could eliminate the `filesystemPomPath == null` case. What do you think?

jpmsilva:
Hi @chkal 
Sorry for the lack of updates, but I had some issues that kept me offline for a few days.
I understand your interest in having POM resolution being the default behavior, and I have adjusted this PR accordingly.
I adjusted the remaining two tests to omit version information from the artifacts, and ran the full test suite to make sure everything is working as intended.
Please check if this is what you envisioned, and let me know if you need any further adjustments!
Thanks

chkal:
Thanks a lot for your help with this. IMO the integration test setup is much simpler now! Merging this now!

Issue #53, opened 2013-09-27
Don't fail if TogglzFilter is registered more than once


chkal:
There seems to be a bug in Glassfish which causes the TogglzFilter to be executed twice.

https://java.net/jira/browse/GLASSFISH-18548

See report here:

https://groups.google.com/forum/?fromgroups#!topic/togglz-users/vM8alPVUUxM

Perhaps such situations shouldn't result in an exception.

jpmsilva:
Hi @chkal 
You may want to close this one since #291 has been merged :smile:

chkal:
Thanks. Seems like I missed this one. :laughing:

PR #291, opened 2018-07-15
Fixes for #53

jpmsilva:
Provides a similar (albeit simplified) approach as Spring's OncePerRequest filter, by adding a marker attribute to the request once the TogglzFilter is executed.
Further executions of the filter check if the marker is present, and will simple execute the chain if so.

Because there is no easily observable behavior to test RequestListener failures (org.togglz.servlet.CompositeRequestListener catches all errors and simply logs them), a test is provided that ensures that the marker is added to the request.

Please let me know if you would like to change the behavior of CompositeRequestListener to allow hard failures, and provide a stronger test case.


chkal:
Thanks a lot! I'll review the pull request in detail as soon as possible. But I'm quite busy at the moment, so it may take some days. Sorry about that.

servlet/src/test/java/org/togglz/servlet/test/util/HttpServletRequestHolderTest.java
```diff
@@ -24,9 +25,12 @@
     public static WebArchive createDeployment() {
         return Deployments.getBasicWebArchive()
             .addClass(HttpServletRequestHolderServlet.class)
+            .addClass(TogglzFilterServlet.class)
             // we don't need to bootstrap Togglz here as we only test the request holder
             .setWebXML(Packaging.webAppDescriptor()
                 .contextParam("org.togglz.FEATURE_MANAGER_PROVIDED", "true")
+                .filter(TogglzFilter.class, "/*")
```
chkal: Could you add a comment here to explain why the filter is added twice in the test! Just to make sure that nobody removes the duplicated registration by accident in a few years. :laughing:

servlet/src/test/java/org/togglz/servlet/test/util/HttpServletRequestHolderTest.java
```diff
@@ -24,9 +25,12 @@
     public static WebArchive createDeployment() {
         return Deployments.getBasicWebArchive()
             .addClass(HttpServletRequestHolderServlet.class)
+            .addClass(TogglzFilterServlet.class)
```
chkal: I'm wondering if the test wouldn't be easier, if we simply have a single Servlet for the test. If `HttpServletRequestHolderServlet` would also output the value of the `EXECUTED` attribute, we would only have a single servlet. The servlet could output something like:

```
Query: number=42
Executed: true
```

Then we just need to adjust the asserts to check for `Query: number=42` instead of just `number=42`.

jpmsilva: Believe or not, that was exactly how I implemented it the first iteration :smile: 

I wanted to have a better way to test the feature, other than looking at the side effects of the implementation (the request attribute marker), but I could not find a better way without breaking the implied contract of `org.togglz.servlet.CompositeRequestListener`.

Anyway, I'll submit a new version very soon!

chkal:
Thanks a lot! Looks great! I'll merge this now.

FeuTeston:
@chkal hi there. Is it possible to release this change (possibly with the other changes already on master) in a non snapshot version of togglz? I am asking since I am facing the same issue this will resolve. In case it is not possible, is there a plan to create a new version? If so, when? 

Before anything, thanks a million for the efforts. I see you are busy and looking for a maintainer, so needless to say that I understand any delays that might happen. Having an heavily used opensource project is no easy task. If I can assist and provide any help, let me know. :-)

chkal:
Hey @FeuTeston. I just started a last attempt to find a sustainable model which allows me to keep working on Togglz and to ensure the community gets new releases and patches. You can learn more about it in my [recent mail](https://groups.google.com/d/msg/togglz-dev/-DPU5w_slG8/mRNe6Cw9BgAJ) on the mailing list.

PR #295, opened 2018-07-28
Generic enums support

ruifigueira:
First attempt, basically copy and adapting existing code, for instance, `EnumBasedFeatureProvider`.

Check #293 for more details.

### DIFF
--- a/core/src/main/java/org/togglz/core/GenericEnumFeature.java
+++ b/core/src/main/java/org/togglz/core/GenericEnumFeature.java
@@ -1,0 +1,35 @@
+package org.togglz.core;
+
+/**
+ * Feature implementation that wraps generic enumeration classes
+ *
+ * @author rui.figueira
+ */
+public class GenericEnumFeature implements Feature {
+
+    private final Enum<?> enumValue;
+
+    public GenericEnumFeature(Enum<?> enumValue) {
+        this.enumValue = enumValue;
+    }
+
+    @Override
+    public String name() {
+        return enumValue.name();
+    }
+
+    public Enum<?> getEnumValue() {
+        return enumValue;
+    }
+
+    @Override
+    public int hashCode() {
+        return enumValue.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof GenericEnumFeature && enumValue == ((GenericEnumFeature) obj).enumValue;
+    }
+
+}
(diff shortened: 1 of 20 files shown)


BeneStem:
Neat! I will try that with Kotlin today. Thank you so much!

BeneStem:
What can I do to help you merge this PR?

ruifigueira:
@BeneStem, I just added a kotlin module, as well as spring boot support for generic enums. Check my commits.

I still need to test most of the things, for the time being I'm just throwing code to check if kotlin support is actually feasible.

BeneStem:
@ruifigueira Yeah I saw that and thought the same (code throwing). I am currently trying to make your code changes work in my project as well and I am getting good results ;)

ruifigueira:
That's great!

@chkal, maybe we can include kotlin support in 2.7.0? If that's the case, it's a matter of polishing this code, as well as create more JUnit tests (for now, there are a couple things for the generic enum part) and documenting it

BeneStem:
@ruifigueira have a look at https://github.com/togglz/togglz/pull/293#issuecomment-408679245 for my implementation.

I don't think we actually need the Kotlin extension you wrote :)

chkal:
Sure. I would be fine with including this in 2.7.0. As long as it integrates well with the other parts of Togglz. However, I still think that it would be great to get some feedback from the Kotlin guys on whether the issue you are running into is intended or if it is simply a bug that can be fixed.

BeneStem:
@chkal I here you. This feedback from them would be great.

You can find my whole wrapper implementation here https://github.com/e-breuninger/spring-boot-starter-breuninger/tree/master/spring-boot-starter-breuninger-togglz/src/main/java/com/breuninger/boot/togglz

ruifigueira:
@BeneStem, I'm aware that the kotlin extension code is not needed, but adding some extension methods for enum classes or even `Feature` can help people that uses kotlin because they don't need to keep declaring an `isActive` method, for instance.

BeneStem:
Yeah sure. Good thinking :)

I was just happy that you helped me fix my problem and wanted to be at least a bit helpful. :S

ruifigueira:
@BeneStem, but you made me realize that we actually don't need kotlin extension methods, they can pollute other enum objects, and that's not such a good thing.

For now, I'm happy with the `GenericEnumFeatureProvider`, and I think that's something we can easily merge into togglz, and already provides everything you need to use kotlin enums as "features".

Regarding kotlin utilities, I need to think a little bit more.

BeneStem:
Hey there. Yes. That is exactly what I was thinking about ;)

bennetelli:
Hi there,
I think we don't need this PR anymore as we already have a kotlin wrapper included in togglz?

Issue #298, opened 2018-08-08
Togglz-console cannot be disabled by togglz.console.enabled property (Spring-boot 2.0)


akoval:
Created test application( spring-boot 2.0.4 )with following config

```
togglz.enabled=true
togglz.featureEnums=test.FeatureFlags
togglz.featuresFile=file:./features.properties
togglz.console.enabled=false
togglz.console.path=/togglz-console
togglz.console.secured=false
togglz.console.useManagementPort=false
togglz.endpoint.id=togglz
togglz.endpoint.enabled=true
togglz.endpoint.sensitive=true
togglz.cache.enabled=false
```

Expected result: /togglz-console endpoint doesn't work.
Actual result: I see UI page

ruifigueira:
I just tried your configuration, it seems to be a bug. In the `/actuator/condtions` report I get:

```yaml
"TogglzAutoConfiguration.TogglzConsoleConfiguration": [{
  "condition": "TogglzConsoleBaseConfiguration.OnConsoleAndNotUseManagementPort",
  "message": "AllNestedConditions 1 matched 0 did not; NestedCondition on TogglzConsoleBaseConfiguration.OnConsoleAndNotUseManagementPort.OnNotUseManagementPort @ConditionalOnProperty (togglz.console.use-management-port=false) matched"
}]
...
```

I'l check that `TogglzConsoleBaseConfiguration.OnConsoleAndNotUseManagementPort`, it seems to be the cause of our problem.

ruifigueira:
PR #299 should fix this issue. I added a small regression test too.

chkal:
@akoval I just merged this fix provided by @ruifigueira. Travis will build and deploy the new snapshots in a few minutes. So you could give the latest snapshots a try to verify the fix.

See the last section on this page for details: https://www.togglz.org/download.html

akoval:
@chkal thank you

PR #299, opened 2018-08-08
Fix condition preventing console from being disabled

ruifigueira:
`OnConsoleAndNotUseManagementPort` condition was not working as
expected

Fixes: #298


chkal:
LGTM! I just merged this.

PR #302, opened 2018-10-28
Add support for context-aware ActivationStrategies

PSanetra:
This PR adds support for context-aware ActivationStrategies with custom strategy-dependent contexts.

It contains the following changes:

* New interface `ContextAwareActivationStrategy` extends the `ActivationStrategy` interface, deprecates the `isActive(FeatureState, FeatureUser)` method and adds a `isActive(FeatureState, FeatureUser, T context)` method
* FeatureManager has a new `isActive(Feature, ActivationStrategyContexts)` method
    * `ActivationStrategyContexts` maps context-values to `ContextAwareActivationStrategy`-types
* New `StringWhitelistActivationStrategy` enables Features if the provided String-context is part of a comma-separated list

Example with `StringWhitelistActivationStrategy`:

If you want to check if a string `"myContext"` is whitelisted you can do that like this:
```java
boolean isActive = featureManager.isActive(
    Features.FEATURE_ONE,
    ActivationStrategyContexts.builder()
        .add(StringWhitelistActivationStrategy.class, "myContext")
        .build()
);
```


chkal:
Thanks a lot for the pull request. I'll try to find some time in the next days to have a deeper look at it.

PSanetra:
Oh, I'm sorry to hear that you are stepped down as maintainer. I hope the project is not dying.

At the moment it is only possible to provide one context object per strategy as I think each strategy may expect a different context. Even if two strategies expect equal contexts, I believe this might happen very likely just by accident and could always change in the future. Since every context-aware strategy might expect a different context object, the user has to provide the strategy class, but I could refactor the code so that you could instead provide a strategy-specific string key or something like that.

PR #305, opened 2018-11-03
Allow @AllEnabled and @AllDisabled on parent classes

### LINKED ISSUE #303: Junit 5 AllDisabled not detected if placed on parent class
It's nice to have support for Junit 5 ... even if it's difficult to find out in the doc.

I''m facing an issue with annotation `@AllDisabled` (same with its conterpart `@AllEnabled`).

Working :
* add the annotation on the test method 
* add the annotation on the class containing the test method 

Not Working :
* add the annotation on any parent (abstract) class of the class containing the test method 

More details :
Let say we have an abstract class for tests called `AbstractParentTest`. It contains some logic and tooling shared with all the tests.
Adding a test class `MyTest` that inherit `AbstractParentTest`.
Now I'd like to disable/enable all the `Feature` for a particular test method.
The failing case is when `@AllDisabled` is placed on `AbstractParentTest` type.
- Features are not disabled
- we have `null` value for a parameter `TestFeatureManager` (on test related methods)

rweisleder:
Fixes #303

### DIFF
--- a/junit5/src/main/java/org/togglz/junit5/AllDisabled.java
+++ b/junit5/src/main/java/org/togglz/junit5/AllDisabled.java
@@ -1,6 +1,7 @@
 package org.togglz.junit5;
 
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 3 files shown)


chkal:
LGTM! Thanks!

PR #310, opened 2018-12-11
Add write operation to change toggle state via Spring Boot Actuator

kschlesselmann:
Resolves: togglz#308

### DIFF
--- a/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
+++ b/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
@@ -64,14 +64,9 @@ public class TogglzEndpoint  {
                 .orElseThrow(() -> new IllegalArgumentException("Could not find feature with name " + featureName));
 
         FeatureState featureState = featureManager.getFeatureState(feature);
+        featureState.setEnabled(enabled);
+        featureManager.setFeatureState(featureState);
 
-        if (featureState.isEnabled() != enabled) {
-            featureState.setEnabled(enabled);
-            featureManager.setFeatureState(featureState);
-
-            return new TogglzFeature(feature, featureState);
-        }
-
-        return null;
+        return new TogglzFeature(feature, featureState);
     }
 }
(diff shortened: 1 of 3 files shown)


spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
```diff
@@ -36,9 +39,14 @@
 @Endpoint(id = "togglz")
 public class TogglzEndpoint  {
 
+    private final FeatureProvider featureProvider;
+
```
ruifigueira: I don't think we need `FeatureProvider`. You can get the list of features with `featureManager.getFeatures()`, like in `getAllFeatures` methods.

spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
```diff
@@ -53,4 +61,23 @@ public TogglzEndpoint(FeatureManager featureManager) {
         Collections.sort(features);
         return features;
     }
+
+    @WriteOperation
+    public TogglzFeature setFeatureState(@Selector String featureName, boolean enabled) {
+        final Feature feature = featureProvider.getFeatures().stream()
```
ruifigueira: We should use `featureManager.getFeatures()` here instead.

spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
```diff
@@ -53,4 +61,23 @@ public TogglzEndpoint(FeatureManager featureManager) {
...
+        FeatureState featureState = featureManager.getFeatureState(feature);
+
+        if (featureState.isEnabled() != enabled) {
+            featureState.setEnabled(enabled);
+            featureManager.setFeatureState(featureState);
+
+            return new TogglzFeature(feature, featureState);
+        }
+
+        return null;
```
ruifigueira: I would prefer if we still returned a `TogglzFeature` even if the state didn't change.

kschlesselmann: I thought I'd be nice to return `null` here because in that case the actuator will respond with HTTP 204 instead of HTTP 200 which enables the client to deal with no actual changes (if someone would need that).

If you still prefer to return something else let's discuss that. In order to provide a more advanced HTTP response we need to add a `@EndpointWebExtension` I think.

kschlesselmann: OK, now the endpoint always performs the set feature state operation and returns its result.

PR #319, opened 2019-07-05
Exchanging Response code 403 Forbidden to 400 Bad Request

joshuabacher2992:
this makes it easier to distinguish errors caused by wrong requests against authentification issues.


bennetelli:
@guidosteinacker1969 I would like to merge this request. A Http 400 instead of 403 seems to be more correct from a semantical point of view. What's your opinion?

guidosteinacker1969:
Sure, should be 400 - and as there should not be any compatibility-issues...

PR #327, opened 2019-10-08
Remove GitHub template files

bennetelli:
We don't need these templates anymore because we have active maintainers :-)


bennetelli:
> > We don't need these templates anymore because we have active maintainers :-)
> 
> Is that true? If it is, that's great! But I'm not seeing any posts/announcements regarding new maintainers (vs casual contributors).
> 
> I've checked the github page, the homepage, the google group, and the Patreon page, and none mention a new maintainer(s) or Christian coming back onto the project?
> 
> Could you elaborate? Thanks!
> 
> -DF

It is :) Will prepare a post about it within the next days. Didn't have time to write about it yet. 
I will maintain the project together with two colleagues.

PR #329, opened 2019-10-10
Add OWASP dependency check plugin

### LINKED ISSUE #328: Add OWASP Dependency check plugin
It would be nice to check our projects as often and early as possible to avoid security issues.

https://jeremylong.github.io/DependencyCheck/dependency-check-maven/

aprueller:
Added OWASP dependency check plugin and suppressed false positive findings.
Currently the integration is not failing builds but generating one aggregate report in `target/dependency-check-report.html`. Before thinking about using automatically breaking builds we should first fix all existing findings.

I currently have no clue how to make the results from the Travis build available in some human readable way. The OWASP dependency check plugin generates a quite useful HTML report but I've not found a way yet to make accessible in Travis :-/.

Fixes #328

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: java
+cache:
+  directories:
+    - $HOME/.m2/repository/org/owasp/dependency-check-data/
 install:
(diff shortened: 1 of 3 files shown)


bennetelli:
Thanks for your PR @aprueller :-)

PR #340, opened 2019-11-05
Updates to Spring Boot 2.2

guidosteinacker1969:


bennetelli:
@guidosteinacker1969 your update breaks our togglz-samples build. We should fix it before merging this PR.

PR #349, opened 2019-11-21
Add reactive starter

bennetelli:
we need an additional starter for reactive WebFlux applications without Thymeleaf, the Togglz Console and corresponding dependencies.

### DIFF
--- a/spring-boot/pom.xml
+++ b/spring-boot/pom.xml
@@ -20,6 +20,7 @@
   <modules>
     <module>legacy-starter</module>
     <module>starter</module>
+    <module>reactive-starter</module>
   </modules>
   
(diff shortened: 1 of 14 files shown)


tdevilleduc:
Has it been merged ? if not, does toggle can be reactive with new version ?

bennetelli:
Hey @tdevilleduc i haven't merged it so far, but I fixed an issue with the togglz actuator endpoint in togglz 2.9.3 and a colleague of mine told me he would be able to work with togglz since that fix. 
Do you still need a separate reactive starter? If yes, why do you need it? Because of the togglz console? Would be great to let me know so I can work on additional features that would be helpful for you.

Issue #354, opened 2019-12-04
HazelcastStateRepository does not allow for client-server Hazelcast model


Jeffrey-Hassan:
The HazelcastStateRepository takes a Config object only, which is the configuration for Hazelcast server.  The constructors + builders should also allow for using:

* ClientConfig for client-server model and 
* HazelcastInstance when we'd rather build that externally (can be built with either client or server configuration)

Therefore there should be 3 constructors in addition to the builder.  Additionally, builder should throw exception if more than 1 of the 3 (Config, ClientConfig, HazelcastInstance) is configured.

bennetelli:
fixed in PR #355

PR #355, opened 2019-12-04
Adding HazelcastInstance plus ClientConfig for HazelcastStateRepository

Jeffrey-Hassan:


Jeffrey-Hassan:
Attempt at resolving Issue #354 .  Use case is I have an application with Hazelcast set up in a client-server model, and would rather just pass in the HazelcastInstance than have the library take care of building one for me.  While I was at it, figured it'd be a good feature to also allow a ClientConfig to be passed in, in case someone wanted to use that as well.

bennetelli:
@Jeffrey-Hassan thanks for opening the issue and thanks for your pull request as well. Let me know if you need additional features or stuff like that. Have a nice day :)

PR #359, opened 2019-12-11
Lazy initialization of instance collections

bhattacharyyasom:
Making the instance collections lazy. Considering there could be potentially numerous providers this will make memory usage optimal.
Also, now we throw a IllegalStateException if we do a getMetaData on a empty provider.

### DIFF
--- a/core/src/main/java/org/togglz/core/manager/EnumBasedFeatureProvider.java
+++ b/core/src/main/java/org/togglz/core/manager/EnumBasedFeatureProvider.java
@@ -1,17 +1,11 @@
 package org.togglz.core.manager;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.Map;
-import java.util.Set;
-
 import org.togglz.core.Feature;
 import org.togglz.core.metadata.FeatureMetaData;
 import org.togglz.core.metadata.enums.EnumFeatureMetaData;
 import org.togglz.core.spi.FeatureProvider;
+
+import java.util.*;
 
 /**
  * Implementation of {@link FeatureProvider} that uses an Java enum to represent features.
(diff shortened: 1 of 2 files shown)


bennetelli:
@bhattacharyyasom do you think this will have a huge performance impact? I don't have that much Features during my daily togglz usage and I think the performance impact is not a problem actually. Looking forward to your response :)

sombhattacharyya1983:
Actually yesterday when I made the change even I considered that thought. Just that I wanted to have that conversation. So , there are two things on my  mind. For the collections stuff I just thought this would be following a more performant practice. I agree though that this maybe overkill.
However the other thing that was on my mind is that the `EnumBasedFeatureProvider ` class allows creation of objects using the default constructor. Considering that we can avoid creating two unnecessary collection objects ahead of time. 
The other thing on my mind is if we do allow creating objects without adding features then shouldn't we return some message to indicate that when I call a getFeatures or getMetaData on that object ? 
Can we rethink the design of the class as it stands today ?

bhattacharyyasom:
> @bhattacharyyasom do you think this will have a huge performance impact? I don't have that much Features during my daily togglz usage and I think the performance impact is not a problem actually. Looking forward to your response :)

Actually yesterday when I made the change even I considered that thought. Just that I wanted to have that conversation. So , there are two things on my  mind. For the collections stuff I just thought this would be following a more performant practice. I agree though that this maybe overkill.
However the other thing that was on my mind is that the `EnumBasedFeatureProvider ` class allows creation of objects using the default constructor. Considering that we can avoid creating two unnecessary collection objects ahead of time. 
The other thing on my mind is if we do allow creating objects without adding features then shouldn't we return some message to indicate that when I call a getFeatures or getMetaData on that object ? 
Can we rethink the design of the class as it stands today ?

bennetelli:
@gsteinacker do you have an opinion about it?

bennetelli:
@bhattacharyyasom good point. I agree with your argumentation and would like to merge your PR. I never used it the way you described it, but it could be used like that by others. That's why I think it is a good idea to initialize it lazy and throw an exception in case of such a misconfiguration.

bhattacharyyasom:
Thanks. :).

PR #369, opened 2020-05-12
Verify if node exists in zookeeper

evilkid:
The current implementation assumes that feature key (zkNode) is existing when updating the state.
In case of Zookeeper version 3.x, an exception is thrown by zookeeper if the node is missing.
I believe this should not be the case, and we need to verify if the node exists or not, and create or update the state accordingly.


evilkid:
It appears that I missed configured my POM file dependencies.

PR #393, opened 2020-06-25
TogglzTestExecutionListener - If not enabled by default, state cannot be enabled

kroegeloh:
I would like to propose this change, because in case of the DefaultsFeatureManager the state will be everytime false, although it is configured different. So perhaps this could be a good option. When defaults isEnabled then change the state to enabled and otherwise the state should be as it is, so the explicit configuration of the feature toggle will be used. 

Perhaps it could be an option to delete line 36 because the default from meta data is gets checked in getFeatureState too, if the state of the feature is null. Am I wrong with it?

### DIFF
--- a/spring-core/src/main/java/org/togglz/spring/test/TogglzTestExecutionListener.java
+++ b/spring-core/src/main/java/org/togglz/spring/test/TogglzTestExecutionListener.java
@@ -37,8 +37,6 @@ public class TogglzTestExecutionListener extends AbstractTestExecutionListener {
 			FeatureState state = manager.getFeatureState(feature);
 			if (defaults.isEnabled()) {
 				state.enable();
-			} else {
-				state.disable();
 			}
 			manager.setFeatureState(state);
 		}


kroegeloh:
Hi,
we set the togglz in our application.yml in the tests and the togglz enable or disable feature. So we have a configuration like this
togglz:
  feature-enums: finance.egp.shared.feature.EgpFeatures
  features:
    PI2019_2_ANACREDIT_FINAL.enabled: true
    PI2019_2_RELEASE_FINAL.enabled: true
    PI2019_3_RELEASE_FINAL.enabled: true

So there is for every release such a toggle. I can only enable and disable predefined features. So every FeatureEnum i don't set to true is false. We use the MetaData to have some strings describing the releases and do not use the defaults for the toggles, because they are booleans with are either true or false. The problem i have with the TogglzTestExecutionManager is, that our EnumFeatureMeta.getDefaultFeatureState is false, because we do not set it automatically to the defined FeatureState of the application.yml and so the state gets turned to disabled. I would like to respect the default state only if state is null and actually the defaults state is respected in both cases state == null and state != null. Am I miss out something?

bennetschulz1309:
Ok, good point. I will checkout the branch and will think about it once again, but I think your points are valid. Will probably merge this PR this evening.

Thanks for your explanation :)

bennetelli:
@kroegeloh Thanks for your Pull Request :-) Will build a new release the next days.

PR #413, opened 2020-08-11
Add HeaderActivationStrategy

### LINKED ISSUE #412: Add default ActivationStrategy based on header
Hi,

I was kind of expecting to have a DefaultActivationStrategy for a predefined header that will be processed in the scope of spring-boot with a servlet configuration, similar to eg. the [QueryParameterActivationStrategy](https://github.com/togglz/togglz/blob/master/servlet/src/main/java/org/togglz/servlet/activation/QueryParameterActivationStrategy.java). I saw in #237 that someone else created a custom Strategy for this, so I guess for now I might have to do the same

Is there any plan to implement a header based activation stragtegy?

huehnerlady:
closes #412 

Here I am trying to fix my own problem.
I have implemented this ActivationStrategy in kotlin and find it works exactly as I would want if using Headers as a strategy.
So I thought I would contribute this towards the togglz framework as well.
I am not that familiar with Java anymore, so I would be happy to hear any feedback.

## Changes

With this Pull Request there is another ActivationStrategy added that is parsing the requestHeader `X-Features`. 

So if you would do the call `curl -L -X GET 'http://localhost:8080/foo' -H 'X-Features: FEATURE_1,FEATURE_2' ` you would activate FEATURE_1 and FEATURE_2

### DIFF
--- a/servlet/src/main/java/org/togglz/servlet/activation/HeaderActivationStrategy.java
+++ b/servlet/src/main/java/org/togglz/servlet/activation/HeaderActivationStrategy.java
@@ -1,0 +1,44 @@
+package org.togglz.servlet.activation;
+
+import org.togglz.core.activation.Parameter;
+import org.togglz.core.repository.FeatureState;
+import org.togglz.core.spi.ActivationStrategy;
+import org.togglz.core.user.FeatureUser;
+import org.togglz.servlet.util.HttpServletRequestHolder;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.Arrays;
+
+public class HeaderActivationStrategy implements ActivationStrategy {
+
+    static final String ID = "header";
+
+    @Override
+    public String getId() {
+        return ID;
+    }
+
+    @Override
+    public String getName() {
+        return "Header";
+    }
+
+    @Override
+    public boolean isActive(FeatureState featureState, FeatureUser user) {
+        HttpServletRequest request = HttpServletRequestHolder.get();
+        if (request == null || featureState == null || featureState.getFeature() == null) {
+            return false;
+        }
+        String header = request.getHeader("X-Features");
+        if (header == null) {
+            return false;
+        }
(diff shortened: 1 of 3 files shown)


bennetelli:
Many thanks for your PR :) 

I just remembered that you can also use Spring Boot's actuator endpoint to enable and disable togglz. Not sure if this would also solve your problem?

Will take a look at your PR now :)

huehnerlady:
> I just remembered that you can also use Spring Boot's actuator endpoint to enable and disable togglz. Not sure if this would also solve your problem?
> 

@bennetelli unfortunately this would not help in my case, as we support parallel requests und it could be that there are 2 requests at the same time where one has a feature enabled and one hasn't. so it would not work to enable an endpoint, make a call, and then disable it again as it would enable the feature for all requests in that timeframe

bennetelli:
Okay. Then I would like to merge your PR.

Quite interesting how different the use cases are :) will build a release either today or tomorrow.

huehnerlady:
@bennetelli thank you so much for your help, looking forward to the release :)

bennetelli:
@huehnerlady I have to say thank you to you :) It's your work. I'm just the "merger" ;-) thank you so much for offering your time and implementing it :)

huehnerlady:
> @huehnerlady I have to say thank you to you :) It's your work. I'm just the "merger" ;-) thank you so much for offering your time and implementing it :)

Oh that's not a problem at all, I did implement it anyways, the only time I needed was to convert from kotlin to java :)

I thank you that you accept my proposal so we do not need to reimplement this in our own lib or worse in every service separately :)

huehnerlady:
@bennetelli do you have any update on this? Is there a timeline when the net release will happen? :)

bennetelli:
Sorry Ruth :( will be my first task tomorrow morning. Had to update additional libraries and dependencies as well :(

Issue #237, opened 2017-07-26
Spring Boot integration to auto-configure filter to populate HttpServletRequestHolder  


taasjord:
I am using the Togglz spring-boot-starter and added a custom strategy to check HTTP header parameter for activation (didn't immediately find one out-of-the-box) . But I noticed that HttpServletRequestHolder does not contain any request object, so I rolled my own filter to populate it. But shouldn't auto-configuration do this by default, in either a filter or interceptor, like the vanilla TogglzFilter does? I have not tested but I assume the other default servlet based strategies maybe doesn't work without it either. 

 I found that it registers SpringSecurityUserProvider if spring-security is on the classpath, so I guess username strategy is taken care of, but I think that the ClientIp and ServerName strategies cannot work without it, unless I missed something here.

chkal:
Hey @marceloverdijk, could you share your thoughts on this?

resah:
Hi,
I have the same problem: I wanted to create a Cookie-based AuthenticationStrategy, but I can't retrieve the HttpServletRequest/HttpServletRequestHolder.

@taasjord Could you help me out with an example implementation of your filter/interceptor until this problem is fixed?

Thanks

taasjord:
Hi @resah. 

Here is my servlet filter to populate the HttpServletRequestHolder:
```
    /**
     * Binds the HttpServletRequest to togglz' HttpServletRequestHolder,
     * since the togglz spring boot starter does not seem to do that.
     */
    protected static class HttpServletRequestHolderFilter implements Filter {

        @Override
        public void init(FilterConfig fc) throws ServletException {
        }

        @Override
        public void doFilter(ServletRequest req, ServletResponse res, FilterChain fc) throws IOException, ServletException {
            try {
                HttpServletRequestHolder.bind((HttpServletRequest) req);
                fc.doFilter(req, res);
            } finally {
                HttpServletRequestHolder.release();
            }
        }

        @Override
        public void destroy() {
        }
    }
```

chkal:
So, what's the official way to automatically configure filters in a Spring Boot app? Togglz uses a `web-fragment.xml` file for Servlet containers. I'm surprised that this doesn't work on Spring Boot.

marceloverdijk:
In Boot I would normally use a `FilterRegistrationBean` (see also https://stackoverflow.com/questions/19825946/how-to-add-a-filter-class-in-spring-boot).

[...]

PR #414, opened 2020-08-11
Add HttpServletRequestHolderFilter

huehnerlady:
As an alternative to the [HttpServletRequestHolderListener](https://github.com/togglz/togglz/blob/master/servlet/src/main/java/org/togglz/servlet/util/HttpServletRequestHolderListener.java) here a Filter that does add the Request to the HttpServletRequestHolder.

In my opinion this would also help with the issue #237 

I implemented this for our use in kotlin and thought I offer the implementation to the community. 
Happy for feedback :)


huehnerlady:
I need help here as it seems coincidental that the build failed(connection reset) and so would like to rerun the check, but do not have a travis account :)

bennetelli:
@huehnerlady Ruth, is this PR important for the other PR I already merged?

huehnerlady:
@bennetelli sorry For some reason I did not get a notification.

This  PR is separate but I thought I share my code as it is similar to the Listener but for Spring-boot and it was asked in an issue before :)

PR #441, opened 2020-10-09
Fix togglz-kotlin artifactId and update version

deen13:
This pull request adds the missing togglz- prefix to the artifactId and bumps the version to latest.


bennetelli:
@deen13 Thanks for your PR :+1:

Issue #455, opened 2020-11-01
Warning during Spring Boot startup with dedicated management port


fabianlinz:
When using Togglz Spring Boot starter (`org.togglz:togglz-spring-boot-starter:2.8.0`) and Spring Actuator (`org.springframework.boot:spring-boot-starter-actuator`) on a separate port (`management.server.port=8081`)
`TogglzApplicationContextBinderApplicationListener` logs the warning : `ApplicationContext already bound to current context class loader, releasing it first`.

#279 introduced a handling for this setup in the `TogglzEndpointAutoConfiguration`. But due to the definition in `spring.factories` this auto configuration is currently not picked up automatically 
(only `TogglzAutoConfiguration` is picked up automatically for auto configuration)

Workaround: Add `@ImportAutoConfiguration(TogglzEndpointAutoConfiguration.class)`

fabianlinz:
Fixed in 2.9.0

Issue #429, opened 2020-09-10
Spring Boot Actuator "togglz" endpoint not exposed


bennetelli:

deen13:
I ran into this issue also while setting up togglz in our kotlin project. For some reason the actuator endpoint is exposed if I convert the `com.ctrace.wallboard.togglz.TogglzEndpointAutoConfiguration` into a kotlin class.

bennetelli:
@deen13 thanks for your response. I remember working on it some months ago. Will take a look at it.

alfonz19:
there seems to be duplicate key in 2.8.0/togglz-spring-boot-starter-2.8.0.jar!/META-INF/spring.factories

fabianlinz:
This should be fixed with 2.9.0

bennetelli:
@fabianlinz thanks for the hint. Indeed. It's fixed already. So I would like to close this issue.

PR #456, opened 2020-11-01
fix Spring Boot auto configuration declaration to include TogglzEndpointAutoConfiguration (in addition to TogglzAutoConfiguration) (#455)

fabianlinz:
This should also solve #429


fabianlinz:
#455

bennetelli:
Thanks for your PR @fabianlinz :)

rhanton:
Sorry, as someone trying to use togglz with a spring boot 1.x application, the fact that this was fixed in v2.9.0 at the same time support for spring-boot 1.x was removed is a pain. I'm going to try extending TogglzAutoConfiguration in my application to get it to run. 🤞

hennr:
Sorry, yes, you have to.
We can't support multiple versions of spring-boot.

bennetelli:
hm. Today Spring Boot 3 is going to be released. I think it's actually time to migrate from Spring Boot 1 to 2 at least. Sorry :(

PR #478, opened 2021-01-06
No restriction of @FeaturesAreActive for enum-feature-classes

m-schroeer:
Extends the usage of the great feature introduced by @ractive (https://github.com/togglz/togglz/pull/166#issue-74804051).

Since the feature names had to be present as string literals (even before this change) the usage of enum-feature-classes is an unnecessary limitation of @FeaturesAreActive.

To 'fix' this limitation these changes do the following:
- FeaturesAreActive.java:
    - remove annotation  field featureClass()
- FeatureInterceptor.java:
    - remove all enum-related code
    - the check is done by:
        1. exception if not all features-string of annotation are actual features (independently from 'active' status)
        2. check if all features of annotation are active (by FeatureManager)
        3. decide if error response or not

---------
This means you can do the following:
```java
 @Controller
 @RequestMapping("/new-feature")
 @FeaturesAreActive(features="NEW_FEATURE")
 public class MyNewFeature {
     @RequestMapping(method = RequestMethod.GET)
     public String newFeature() {
         return ....;
     }

     @FeaturesAreActive(features="NEW_SECURE_FEATURE", responseStatus=403)
     @RequestMapping(value="/secure", method = RequestMethod.GET)
     public String newSecureFeature() {
         return ....;
     }
 }
```
 
Note the missing `featureClass=MyFeatures.class` in `@FeaturesAreActive(...)`.

### DIFF
--- a/shiro/pom.xml
+++ b/shiro/pom.xml
@@ -35,7 +35,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.12</version>
+      <version>4.13.1</version>
       <scope>test</scope>
     </dependency>
 
--- a/spring-web/src/main/java/org/togglz/spring/web/FeatureInterceptor.java
+++ b/spring-web/src/main/java/org/togglz/spring/web/FeatureInterceptor.java
@@ -5,9 +5,13 @@ import org.springframework.web.method.HandlerMethod;
 import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
 import org.togglz.core.Feature;
 import org.togglz.core.context.FeatureContext;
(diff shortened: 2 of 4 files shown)


m-schroeer:
Maybe some more of my thoughts:
For me it is no big deal to work around this one and implement it by myself without being a feature of Togglz itself.
I just didn't saw the reason for this limitation and thought it could be useful for someone else.
If there's any reason I've overlooked please tell me.

Best regards!

bennetelli:
@m-schroeer I think it's time to say thank you for your thoughts and especially for your work on togglz. Thank you very much :-) I really enjoy reading your thoughts and getting your PRs. Keep going please :)

PR #480, opened 2021-01-06
Re-add 'spring-boot-configuration-processor'

m-schroeer:
Re-add 'spring-boot-configuration-processor' to 'togglz-spring-boot-starter'.

This dependency was removed with former togglz/spring-boot/autoconfigure project during re-structuring of spring-based-projects after 2.6.1.Final.
It is necessary for the IDE binding between files with `@ConfigurationProperties` and `application.properties` file (e. g. [TogglzProperties.java](https://github.com/togglz/togglz/blob/master/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzProperties.java) ) and therefor quite useful (e. g. auto-completion in `application.properties`)

See: https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html#configuration-metadata-annotation-processor

### DIFF
--- a/spring-boot/starter/pom.xml
+++ b/spring-boot/starter/pom.xml
@@ -44,6 +44,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-configuration-processor</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
       <groupId>org.thymeleaf</groupId>
       <artifactId>thymeleaf</artifactId>
       <version>3.0.11.RELEASE</version>


bennetelli:
@m-schroeer thanks for your PR. I didn't need this mapping so far, but we should add it if it's useful for others.

m-schroeer:
> @m-schroeer thanks for your PR. I didn't need this mapping so far, but we should add it if it's useful for others.

I was just experimenting with Togglz and IntelliJ Ultimate (one reason is this great feature) and switched from 2.6.1.Final to newer versions and was confused why this binding and auto-completion stopped working. After a while I just noticed the "lost" dependency.

Best regards!

PR #482, opened 2021-01-06
it's not necessary to set 'register-feature-interceptor: true' proper…

### LINKED ISSUE #481: @FeaturesAreActive is not ready to use by default
If `togglz.web.register-feature-interceptor=true` is not set in `application.properties` the [FeatureInterceptor.java](https://github.com/togglz/togglz/blob/master/spring-web/src/main/java/org/togglz/spring/web/FeatureInterceptor.java) is not auto-configured.

In my opinion this should be the case.

The reason is this missing property `matchIfMissing = true`. (Analog to the change in #479 .

https://github.com/togglz/togglz/blob/7345ae5851cec89d7c15e3fb76283a1153216855/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzAutoConfiguration.java#L231

m-schroeer:
…ty anymore

Fixes #481

### DIFF
--- a/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzAutoConfiguration.java
+++ b/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzAutoConfiguration.java
@@ -228,7 +228,7 @@ public class TogglzAutoConfiguration {
     @Configuration
     @ConditionalOnWebApplication
     @ConditionalOnClass(HandlerInterceptorAdapter.class)
-    @ConditionalOnProperty(prefix = "togglz.web", name = "register-feature-interceptor", havingValue = "true")
+    @ConditionalOnProperty(prefix = "togglz.web", name = "register-feature-interceptor", havingValue = "true", matchIfMissing = true)
     protected static class TogglzFeatureInterceptorConfiguration implements WebMvcConfigurer {
         @Override
         public void addInterceptors(InterceptorRegistry registry) {


bennetelli:
@m-schroeer why do you think this should be true/enabled by default? Is it common usecase?

m-schroeer:
> @m-schroeer why do you think this should be true/enabled by default? Is it common usecase?

I don't know if it is a common use case. I was experimenting with Spring and Togglz and found this feature by looking into the code. After this it took me some time to notice why it didn't worked out of the box. Maybe it is more an issue of documentation (I know, maintainer changed. I am glad that you are continuing this project).

I just didn't saw a reason at first why it should not be enabled by default. Maybe time concerns, because it runs before every request.
For me and our project this feature seems to be a good way to enable/disable plugin-like modules that are providing additional REST-endpoints. But we are still evaluating.

Now that I know about the configuration of this it is no problem for me to leave it as it is and reject this PR.

------
Finally there seems to be an inconsistency in JavaDoc:

https://github.com/togglz/togglz/blob/7d1db1fb2db76ef5ab691b5a5f1ec36e7deddf0c/spring-web/src/main/java/org/togglz/spring/web/FeaturesAreActive.java#L34-L36

vs.

https://github.com/togglz/togglz/blob/7d1db1fb2db76ef5ab691b5a5f1ec36e7deddf0c/spring-web/src/main/java/org/togglz/spring/web/FeatureInterceptor.java#L20

Maybe the JavaDoc in `FeaturesAreActive.java` could be more clear by adding something like the following:
```Java
* [...] This is automatically done by the TogglzAutoConfiguration config which is loaded if @EnableAutoConfiguration
* is used in your project and setting togglz.web.register-feature-interceptor to {@code true}.)
```

bennetelli:
I never experienced performance issues with togglz in the years I'm working with it. So let's go with this pull request as well.
In case someone else gets performance problems, we can also easily change it back from enabled to disabled by default.

PR #489, opened 2021-01-07
update spring and spring-boot version (5.3.2 and 2.4.1)

m-schroeer:
Updated spring versions.

Additionally replaced `HandlerInterceptorAdapter` that was deprecated since `5.3`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
     <!-- Wildfly version used for integration tests -->
     <wildfly.version>8.2.1.Final</wildfly.version>
 
-    <spring.version>5.2.7.RELEASE</spring.version>
+    <spring.version>5.3.2</spring.version>
     <junit.version>5.7.0</junit.version>
   </properties>
 
--- a/spring-boot/starter/pom.xml
+++ b/spring-boot/starter/pom.xml
@@ -72,12 +72,24 @@
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-config</artifactId>
       <optional>true</optional>
+      <exclusions>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-core</artifactId>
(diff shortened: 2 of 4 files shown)


m-schroeer:
The error that let the checks fail is the following:
```
java.lang.NoClassDefFoundError: org/springframework/core/metrics/ApplicationStartup
```
It seems that somewhere is a lack of support for `spring-boot` `2.4.x`(https://stackoverflow.com/questions/65046056/spring-boot-classnotfoundexception-org-springframework-core-metrics-applications).

Sorry for the hasty PR.
Hopefully I find time to research this.

m-schroeer:
The problem seems to be that maven uses an older version of `sporing-core`than `5.3.x`
 and therefor cannot find the class `ApplicationStartup`that was introduced in `5.3.x`.

More specific it uses `spring-core:5.2.11.RELEASE` as you can see in following image:
image

The only place I found where a version prior `5.3` is used is `togglz-spring-security`.
The strange thing is that its configured like this:
https://github.com/togglz/togglz/blob/7d1db1fb2db76ef5ab691b5a5f1ec36e7deddf0c/spring-security/pom.xml#L21-L26

- The set version uses `spring-core:5.2.6.RELEASE` (and not the conflicting `5.2.11.RELEASE`)
   - FYI: If `spring-security-web`would be set to version `5.4.2`it would use the conflicting `spring-core:5.2.11.RELEASE`
- The scope is provided and should not be used anyway for compile phase

I am sure I am overlooking something. I will have to look again...

bennetelli (approved):

bennetelli (approved):

bennetelli (approved):

bennetelli:
@m-schroeer I think this is the way to go. I would also exclude it. In my opinion the spring core dependency is more important. Therefore we should go for it.

PR #561, opened 2021-04-30
Upgrade hazelcast major version

jensim:
# Upgrade hazelcast major version
_Now at 4.2_

https://hazelcast.com/blog/hazelcast-imdg-4-0-is-released/
> A key thing to note is that the client protocol has also been updated in 4.0. This means that you cannot use a 3.x client with 4.0 members. This is something that only changes on a major release. Once you upgrade your clients to 4.x then you can continue to use them with various 4.x Member versions.

The client lib is now included in the core lib, and no longer needed.
The package of IMap has moved from core to map, not much else is new, apart from compatibility.

### DIFF
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -13,7 +13,7 @@
 	<description>Togglz - Hazelcast integration</description>
 
 	<properties>
-		<hazelcast.version>3.12.9</hazelcast.version>
+		<hazelcast.version>4.2</hazelcast.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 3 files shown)


jensim:
I didn't find a contribution markdown in the repo, so I just did a simple PR, hope that's fine by you

I was not able to use the togglz-hazelcast state repo, due to major version indifference, and the package change of IMap, had it not been for that, I would probably have been override the version using maven properties.

bennetelli:
Many Thanks for your PR @jensim :-)

bennetelli:
Of course it's fine for me. Will release a new version today so you can use it again.

PR #595, opened 2021-07-21
Replace JedisPool with Pool<Jedis> 

aschepp:
To allow the usage of JedisSentinelPool inside RedisStateRepository, the used class for storing the pool has to be changed.

Both JedisPool and JedisSentinelPool are subclasses of Pool<Jedis>.


bennetelli (approved):

bennetelli:
@aschepp will release it on monday morning CEST. Thanks for your PR and have a nice weekend :-)

PR #597, opened 2021-07-29
Move MyFeatures to java sourceSet

leonard84:
This should fix the issue, at least it works locally for me.


bennetelli:
Thanks for fixing the issue @leonard84 :-)

Issue #582, opened 2021-05-21
Update nashorn engine to graalvm js


bennetelli:
needs to change:
* "JavaScript" to "graal.js" in the enginename
* add  org.graalvm.js:js dependency
* add  org.graalvm.js:js-scriptengine dependency

bennetelli:
https://stackoverflow.com/questions/58996429/access-java-object-from-javascript-in-graalvm-polyglot-context

bennetelli:
https://golb.hplar.ch/2020/04/java-javascript-engine.html

bennetelli:
finished with PR #600

PR #598, opened 2021-08-02
Build multiple Java versions

bennetelli:


.github/workflows/maven.yml
```diff
@@ -4,15 +4,20 @@ on: [push, pull_request]
 
 jobs:
   build:
-
+    strategy:
+      fail-fast: true
```
hennr, line 8: if you'd make this false

.github/workflows/maven.yml
```diff
@@ -4,15 +4,20 @@ on: [push, pull_request]
...
+      matrix:
+        jdk: [8, 11, 14, 16]
```
hennr, line 10: and if you make this this 8, 11
you should see the build pass now and the java 17 support could go to the graalvm ticket then: #582

hennr:
@bennetelli Is this rebased against the master / contains my last cassandra change?

bennetelli:
> @bennetelli Is this rebased against the master / contains my last cassandra change?

@hennr now I did. But now there is a test failing in slack integration. I tested it locally and it's working on my machine... argh.

bennetelli:
> > @bennetelli Is this rebased against the master / contains my last cassandra change?
> 
> @hennr now I did. But now there is a test failing in slack integration. I tested it locally and it's working on my machine... argh.

ah. the push job is green, but the pull job not. What do you think, is it ok to merge, @hennr ?
I would give it a try and merge it.

hennr:
> > > @bennetelli Is this rebased against the master / contains my last cassandra change?
> > 
> > 
> > @hennr now I did. But now there is a test failing in slack integration. I tested it locally and it's working on my machine... argh.
> 
> ah. the push job is green, but the pull job not. What do you think, is it ok to merge, @hennr ? I would give it a try and merge it.

Well, works on my machine as well...

Agreed, merge it and take it from there if it fails regularly. It never failed on my machine so far at least.

bennetelli:
@hennr looks like it's working: https://github.com/togglz/togglz/commit/b406a6112eee65bc7a2e99e5def9e2faac95ac85

hennr:
That a quite pleasant piece of news 👍🏼

hennr:
@bennetelli what's the deal with build on push and pull_requests btw.? I don't get it fully.
It will build on pull requests, and commits to master and commit to all other branches, that don't are a PR yet, right?

Bad thing is the duplictate build on PRs then, I think.
Would it be ok to build only PRs and commits to the master branch instead?

bennetelli:
Me neither. I also tried to separate both, but that was horrible as well..

yes, probably.

that's the reason why i wanted to get rid of it in he past because of the duplicated runs.

bennetelli:
> Would it be ok to build only PRs and commits to the master branch instead?

I think we can give it a try. You can add it if you want :)

hennr:
ok :)
#673

PR #627, opened 2021-12-03
fix label creation for annotation feature group when no @Label exists

ewoelfel:
hey, I've found a bug when having a feature group with no Label-Annotation.
When having a FeatureGroup like the following:

```
 @FeatureGroup
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @interface ClassLevelUnlabeledGroup {
 }
```
The label was always "Class"

The behavior can be reproduced by the test i added.


ewoelfel:
issue: #628

bennetelli (approved):

bennetelli:
oh, you are right. Thanks for your PR, @ErikWoelfel2323 :)

Issue #626, opened 2021-11-30
togglz-spring-boot-starter incompatible with Spring Boot 2.6


mihkelmuur:
After update to Spring Boot 2.6 an application relying on `togglz-spring-boot-starter` fails to start:
```
Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
2021-11-23 12:55:09.330 ERROR 17296 --- [           main] o.s.b.d.LoggingFailureAnalysisReporter   : 

***************************
APPLICATION FAILED TO START
***************************

Description:

An attempt was made to call a method that does not exist. The attempt was made from the following location:

    org.togglz.spring.boot.actuate.autoconfigure.TogglzManagementContextConfiguration$TogglzConsoleConfiguration.getContextPath(TogglzManagementContextConfiguration.java:62)

The following method did not exist:

    'org.springframework.boot.actuate.autoconfigure.web.server.ManagementServerProperties$Servlet org.springframework.boot.actuate.autoconfigure.web.server.ManagementServerProperties.getServlet()'
...
```

The [Javadoc on `ManagementServerProperties.Servlet.getContextPath()`](https://docs.spring.io/spring-boot/docs/2.5.7/api/org/springframework/boot/actuate/autoconfigure/web/server/ManagementServerProperties.Servlet.html#getContextPath--) suggest to replace `ManagementServerProperties.getServlet().getContextPath()` with `ManagementServerProperties.getBasePath()`:
> _**Deprecated.** since 2.4.0 for removal in 2.6.0 in favor of `ManagementServerProperties.getBasePath()`_

hennr:
I had a look into this and the method together with the corresponding property (to set a prefix path for mangement tooling like actuator etc.) has changed from 2.4.0:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.4.0-Configuration-Changelog
key: management.server.servlet.context-path
replacement: management.server.base-path

As pointed out by @mihkelmuur 2.6.0 dropped the old config key suppor and as such the starter breaks.

This calls for a new togglz release which includes this breaking change: Spring-Boot < 2.4.x won't work with this new release.

hennr:
I opened a PR to fix this.
Feedback and reviews are welcome.

hennr:
FYI @PH1P5

hennr:
Thanks for merging, @bennetelli , my pleasure to provide a fix.

I wasn't aware of any sample project until you mentioned them in the PR.
[...]

PR #637, opened 2022-01-21
make spring-boot-starter compatible with spring-boot 2.6x+ (and break…

hennr:
… with < 2.4.x) by adopting the new property key for management server base paths / prefixes. (management.server.servlet.context-path -> management.server.base-path)

Also the build didn't run any more test, as it seems, see:
https://github.com/togglz/togglz/runs/4324826674?check_suite_focus=true#step:4:3788

I removed the JUnit version numbers from the module to inherit from the parent pom and the tests work fine again now.

Would be great if we could have a release afterwards to make toggle work with spring-boot 2.6.

### DIFF
--- a/spring-boot/starter/pom.xml
+++ b/spring-boot/starter/pom.xml
@@ -13,7 +13,7 @@
   <description>Togglz - Spring Boot Starter</description>
 
   <properties>
-    <spring-boot.version>2.4.1</spring-boot.version>
+    <spring-boot.version>2.6.2</spring-boot.version>
   </properties>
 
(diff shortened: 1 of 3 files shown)


hennr:
This PR fixes #626

hennr:
@bennetelli or any other maintainer, could you approve this PR to have a pipeline run, please?

mle-idealo:
I confirm the issue and also its fix!

bennetelli:
Thanks for your work @hennr! Will update the sample projects as well.

PR #652, opened 2022-02-01
enforce to run existing unit tests to avoid green build for possiby broken, but not executed, tests

hennr:
Hi @bennetelli 
I played a bit more with the project and noticed, already last time, that some unit tests are not executed.

This PR will break the build if that should happen again. It also fixes this issue for the console module,

It also uncovers the problem for the CDI module, I was unable to fix it though. Should I open an issue for that?

I could do the same work for the ITs / failsafe-plugin if you are merging this PR.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -69,6 +69,14 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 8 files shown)


hennr:
superseeded by #653

bennetelli:
Thanks :) will take a look at it later

PR #664, opened 2022-02-07
drop usage of commons lang for a string null / empty check only

hennr:


cassandra/src/main/java/org/togglz/cassandra/CassandraStateRepository.java
```diff
@@ -143,7 +142,7 @@ public void setFeatureState(final FeatureState featureState) {
     }
 
     private void putOrDelete(final ColumnListMutation<String> mutation, final String column, final String value) {
-        if (StringUtils.isBlank(value)) {
+        if (value == null || value.trim().isEmpty()) {
```
bennetelli, line 145: good catch! Thanks! I hate using libraries for things like that :)

bennetelli (approved):

cassandra/src/main/java/org/togglz/cassandra/CassandraStateRepository.java
hennr: same here ;)

PR #665, opened 2022-02-07
slack junit 4 -> junit 5

hennr:
I updated the freshly migrated spock -> junit tests to junit 5


bennetelli:
thanks for your PR @hennr :) I removed the public access modifiers and renamed some test methods as well.

PR #672, opened 2022-02-13
add needed jvm opts to run cassandra tests under java11+; see #598

hennr:
Hi @bennetelli , togglz does build properly with this on my machine.
If you'd change fail-fast: true to false I think you should see the java 8 and 11 builds pass in a matrix build.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -51,6 +51,8 @@
             <artifactId>curator-test</artifactId>
             <version>5.2.0</version>
         </dependency>
+        <!--        brauchen wir hier hier nur die api? nur -simple? nur logback?  nur provided? -->
+        <!--        https://repo1.maven.org/maven2/org/slf4j/ -->
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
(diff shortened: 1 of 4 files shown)


benchmarks/pom.xml
```diff
@@ -51,6 +51,8 @@
             <artifactId>curator-test</artifactId>
             <version>5.2.0</version>
         </dependency>
+        <!--        brauchen wir hier hier nur die api? nur -simple? nur logback?  nur provided? -->
+        <!--        https://repo1.maven.org/maven2/org/slf4j/ -->
```
bennetelli, line 55: good question. I don't know. You can give it a try if you want :)

bennetelli (approved):

bennetelli:
Many thanks for your PR :) I always wanted to upgrade the cassandra dependencies to achieve this. Good that's also possible without upgrading. Nice work!:)

benchmarks/pom.xml
hennr: Whoops, these logging changes were not planned to be in this PR as well ;)

Sorry for that, I guess I have to dig into this as my next todo then :)

bennetelli:
Works with java 8 and 11 on my machine.

bennetelli:
@hennr now I (or we :-) ) need to replace the jdk internal nashorn engine to be able to build jdk 13+

hennr:
> @hennr now I (or we :-) ) need to replace the jdk internal nashorn engine to be able to build jdk 13+

Yeah, well. Let's see 😏

PR #679, opened 2022-02-16
move assertJ to test scope

hennr:
build works fine with assertJ in test scope, so I assume the `<scope>test</scope>` was left out by misstake?


bennetelli:
yes. i moved dependencies around and missed it somehow. thanks

PR #680, opened 2022-02-17
Update README.md

hennr:
http -> https link
drop contributors section, it's integrated on the startpage already and openCollective page is inactive anyways
add logo
let CI badge link point to actions tab

(not sure about the logo link, the png could go to the repo itself as well in favour of using the togglz.org/ link)


hennr:
The easiest way to have a look at the changes is the rendered view here:
https://github.com/hennr/togglz/tree/patch-1

PR #691, opened 2022-02-24
use slfj4-api as the default logging facade and drop JUL support as s…

### LINKED ISSUE #682: migrate completely to slf4j, dropping the togglz logging wrapper
togglz brings it's own wrapper for logging in the core module.
This will look for slf4j and JUL in the classpath and, if both are there, prefer JUL over slf4j.

Some modules don't use the togglz logging wrapper at all though, namely zookeeper and  dynamodb.

Slf4j is the defacto standard and defaulting to JUL seems outdated to me.
I propose to drop the inhouse logging wrapper (thus the slf4j module, and the SPI stuff around the LogProvider) and migrate all logging to the slf4j api, leaving it to the consumers to provide an implementation like logback during runtime.

I also suggest to using logback for all test logging in combination with a sane logback-test.xml for less verbose, and thus hopefully faster, test runs.

Any opnions on this, @bennetelli ?

hennr:
…uch. Also use logback-classic for test logs. Will close #682


hennr:
This is quite a biggy, I will add some more words about what's happening here during the next days.
You could scan changes and build logs already though.

Also it would make sense to test this branch integrated in an actual application to see the logging behaviour there. I didn't do this yet.

bennetelli:
we could add it here: https://github.com/togglz/togglz-samples

hennr:
To the changes in this PR:

There are two major changes made:

## drop inhouse log configuration
togglz has it's own togglz/core/logging/LogFactory which can be used to write logs. There where some modules making use of it (e.g. core) and there where some other that used slf4j (e.g. dynamodb) right away.
The home brewn logging setup allowed to switch between JUL and SLF4J during runtime, but as just said it's usage was inconsitent accross the modules.
This PR removes that configurability and uses slf4j-api as a dependency instead. Thus  the slf4j module got purged completely.
Consumers can then use logback or slf4j-simple to see togglz log messages. If they don't have it in the classpath, logs will automatically be ommitted.
Even if consumers want to stick to JUL, slf4j has support for this as well: https://search.maven.org/artifact/org.slf4j/slf4j-jdk14
I consider this to be a simplification as such without even dropping features for the users.

## add logback-classic + config for all tests
slf4j is quite widely used and as such I figured it would be a good idea to use logback-classic to see togglz and third party lib logging during test execution time.
As such I added logback-classic with a test scope and added a logback-test.xml for each module under src/test/resources
This allows for fine grained logging output and I choose INFO as the default log level, decreasing log output for `mvn clean verify` by around 2000 lines.
Not all logs make it to logback yet though but the most at least.

We could have a chat about this later on or do it async via comments here.
When this hits the master branch, a new release would be very appreciated also for all the test dependencies that I moved to test scope in the past to get a cleaner production classpath for consumers.

Cheers

bennetelli:
@hennr do you know what's wrong here?

hennr:
Yep, seems to be a race condition.
The sender has not enough time to deliver the slack notification before we check on the receiver. I added a Thread.sleep(500) which should do the trick.
Not that nice, but as there is only one test that needs this, I decided not implement a dynamic wait here to save some more milis.

PR #710, opened 2022-03-25
Integrate togglz samples

bennetelli:
currently WIP

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,7 @@
     <module>slack</module>
     <module>redis</module>
     <module>kotlin</module>
+    <module>samples/spring-boot</module>
   </modules>
 
   <developers>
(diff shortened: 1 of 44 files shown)


hennr:
Hey @bennetelli  Do you just want to have the examples closer to the code or does this increase our test capabilites as well?
If not I would rather stick with the extra repo and probably play around with auto merged PRs for dependency updates to lower the maintenance effort for the examples.

bennetelli:
> Hey @bennetelli Do you just want to have the examples closer to the code or does this increase our test capabilites as well? If not I would rather stick with the extra repo and probably play around with auto merged PRs for dependency updates to lower the maintenance effort for the examples.

I do have a code snippet for auto merged dependabot PRs on my disk. I would like to add this independent from the samples project.

My intention was more like that I would like to see immidiately when I break things with the samples when I change the togglz code. Right now I would have to update the samples projects by hand every time I release a new version of togglz. 
I just thought by adding this as a module like every other module (e.g. guice) to togglz, so this will be build every time I change something.

bennetelli:
already covered in #710

PR #726, opened 2022-04-07
Leave inner space of parameter values unchanged

paweusz:
Having inner space in parameters untouched allows using space as separator inside multi-valued parameters. Good example could be ClientIpActivationStrategy where one can provide a list of IPs (space separated) as value of parameter.

### DIFF
--- a/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
+++ b/spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/TogglzEndpoint.java
@@ -29,14 +29,10 @@ import org.togglz.core.repository.FeatureState;
 import org.togglz.core.util.Preconditions;
 import org.togglz.spring.boot.actuate.autoconfigure.TogglzFeature;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import static org.springframework.util.StringUtils.commaDelimitedListToSet;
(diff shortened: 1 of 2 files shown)


hennr:
Hi @paweusz ,

I am not sure what this will do to existing code / usage.
Can't you simply use a comma as a seperator, leaving the rest of the code as is and as such reducing the risk to break things for other users?

paweusz:
Hi @hennr,

The issue with current approach is that comma is already used as a separator for parameters.
If I'd like to provide multiple IP addresses for `ClientIpActivationStrategy` (value of one parameter) separated by comma, actuator endpoint would treat it as separate parameters and report an error. 
The other option is to provide space separated list of IPs which is already handled by `ClientIpActivationStrategy`. But removing of all whitespace from parameter value caused IP addresses to be concatenated into one string. This fix leaves inner space of parameter values untouched and provides a way to deliver multiple IPs to `ClientIpActivationStrategy`.

hennr:
Any opinion on this @bennetelli ?

bennetelli:
That explanation makes sense to me. I would say we can mege this PR as it helps @paweusz and doesn't change much.
I hope I am not wrong, but I expect this to not affect other users use cases and workflows.

bennetelli:
using a semicolon as separator would also work and would also be something users could be expecting. But I am fine with spaces as separator as well. 
To be honest, I don't think there are much users out there using the actator endpoint directly and even if there are lots of users doing this, I would expect they don't have a use case like this.

bennetelli (approved):

bennetelli:
@paweusz I just released a new version. 3.1.2 contains your PR. Thanks for raising it and sorry for the delay.

PR #743, opened 2022-04-24
editorconfig proposal for unified indentention, LF etc.

hennr:
@bennetelli any changes needed from your side?


bennetelli:
> @bennetelli any changes needed from your side?

right now not. We can add additional things later if we want. Thanks for the PR! Let's go for it :)

PR #754, opened 2022-04-27
Add dependabot auto merge workflow

bennetelli:


hennr:
Hi @bennetelli ,

feel free to merge and give it a try!
Not sure about some lines in the new worklfow though, but let's iterate in case this does not work as expected.

bennetelli:
yes, but right now the secrets won't work :) will do it when I have time

Issue #575, opened 2021-05-17
Restore Feature.isActive as a default method


drekbour:
Should togglz restore the `Feature.isActive()` interface method as a default impl?

bennetelli:
It should :)

PR #756, opened 2022-05-04
Add default isActive implementation

wyhasany:
In order to simplify usage of enum Feutures is nice
to have default implementation and write less
boilerplate.

Solves #575

### DIFF
--- a/core/src/main/java/org/togglz/core/Feature.java
+++ b/core/src/main/java/org/togglz/core/Feature.java
@@ -1,37 +1,31 @@
 package org.togglz.core;
 
+import org.togglz.core.context.FeatureContext;
+
 /**
- * 
  * <p>
  * This interface represents a feature and is typically implemented by the feature enum.
  * </p>
- * 
- * <p>
- * Usually it makes sense to implement the following method which allows a very easy way to check the status of a feature.
- * </p>
- * 
- * <pre>
- * public boolean isActive() {
- *     return FeatureContext.getFeatureManager().isActive(this);
- * }
- * </pre>
- * 
- * <p>
- * Please note that in Togglz 2.0 the <code>isActive()</code> method isn't defined in the interface any more. So when updating
- * to Togglz 2.0, you will have to remove the {@link Override} annotation from your implementation of the method.
- * </p>
- * 
+ *
  * @author Christian Kaltepoth
- * 
+ * @author Michał Rowicki
  */
 public interface Feature {
 
     /**
      * Returns a textual representation of the feature. This method is implicitly implemented as feature typically are
      * enumerations.
-     * 
+     *
      * @return Name of the feature
      */
     String name();
 
+    /**
+     * Checks whether the feature is active for the current user.
+     *
+     * @return <code>true</code> if the feature is active, <code>false</code> otherwise
+     */
+    default boolean isActive() {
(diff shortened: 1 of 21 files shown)


core/src/main/java/org/togglz/core/Feature.java
```diff
@@ -1,37 +1,31 @@
...
- * 
- * <p>
- * Please note that in Togglz 2.0 the <code>isActive()</code> method isn't defined in the interface any more. So when updating
- * to Togglz 2.0, you will have to remove the {@link Override} annotation from your implementation of the method.
- * </p>
- * 
+ *
  * @author Christian Kaltepoth
- * 
+ * @author Michał Rowicki
```
bennetelli, line 11: It's no need to add your name. Of course you can do that if you want, but I am not requiring this :) so if you want to get contacted, you can add your name here of course :)

bennetelli:
Many thanks for your PR @wyhasany :) I really appreciate it :) I would like to play around with it a bit and also integrate and update the samples projects before, but then I will merge your PR. Hopefully I will find time today in the evening, but I will merge it for sure until the end of the week. I swear :)

core/src/main/java/org/togglz/core/Feature.java
wyhasany: I'm just get used to it from spring-framework repos, anyway I'll remove it :)

bennetelli: you will be in the hall of fame of committers anyway :)

bennetelli:
@chkal I found a commit/PR where you first "Added isActive() to the Feature interface" and removed it some day later with this commit message "Removed isActive() from the Feature interface". Do you remember why you removed it? I can't find any hints in the code. Would be interesting to know :)

core/src/main/java/org/togglz/core/Feature.java
wyhasany: I'm aware of `git blame` 🤣

bennetelli: haha :-D

drekbour:
I'm guessing, it would have been removed for the simple reason that it's wrong to _force_ that unimplemented method onto all implementations of `Feature` (despite 99% adding it).

bennetelli:
> I'm guessing, it would have been removed for the simple reason that it's wrong to _force_ that unimplemented method onto all implementations of `Feature` (despite 99% adding it).

Could actually be. I would say this would not be that bad in general. It's a breaking change and a bit of work for everyone using togglz, but at least in my experience, users only have one Feature class per application. But I used togglz with Feature enum in Spring Boot most of the time and I could be wrong. I don't really know about the distribution of all possible togglz use cases out there. 

So better ask Christian before. Maybe he can give some insights even after all the years.

wyhasany:
@bennetelli thanks a lot I hope it's not the last contribution here, I'm waiting for release now :)

bennetelli:
> @bennetelli thanks a lot I hope it's not the last contribution here, I'm waiting for release now :)

I hope so, too :) your help is always welcome and appreciated here :) 

i am currently building a release

bennetelli:
@wyhasany @drekbour 3.2.0 has been released. Thanks y'all for your contributions in discussions and code changes :)

PR #758, opened 2022-05-06
Integrate togglz samples in the same repository

bennetelli:


bennetelli:
@hennr 

Right now I am using two different parents:

<parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-parent</artifactId>
    <version>2.6.7</version>
    <relativePath/> 
</parent>

and:

    <parent>
    <groupId>org.togglz</groupId>
    <artifactId>samples</artifactId>
    <version>1.0.0-SNAPSHOT</version>
</parent>
    
From a users perspective I am preferring the standard spring boot one, but I am not sure about the releases then. Do you have an opinion about it?

bennetelli:
I hate Maven.. seriously.

hennr:
Let's have a video call on this :)

bennetelli:
* there is an issue with the fixed java version in the samples
* I fucked something up in togglz 2.9.7 so feature enums don't work the way as before. I am on it. Don't remember yet what I changed there

bennetelli:
this line breaks the feature enums: @ConditionalOnProperty(prefix = "togglz.web", name = "request-holder-filter", havingValue = "true")

hennr:
@bennetelli I won't have time to review this until beginning / mid of next week.
I invested quite some time though to get rid of unneeded dependencies for several togglz modules so I would prefer to make sure we don't end up with more deps to fix afterwards again.

Could this PR wait until our call?

hennr:
Hey @bennetelli,

if we would have the samples in the main repo, we could also use these as integration tests to some extend, right?
I don't see a workflow that would build the samples at the moment, let's discuss this further face to face.

bennetelli:
> Hey @bennetelli,
> 
> if we would have the samples in the main repo, we could also use these as integration tests to some extend, right? I don't see a workflow that would build the samples at the moment, let's discuss this further face to face.

Yes, that's the intention of it. We/I missed the configuration change in spring boot starter because of this separation. That's why I would like to have it "closer" to the project and build and test the samples with the current snapshot.

hennr:
As discussed, feel free to merge when ready.

bennetelli:
MAAAAAAAAAAAN. now I fucked it up. Wenns mal schnell gehen muss..

bennetelli:
the version in the samples projects don't get upgraded. So either we are overriding a released samples project over and over again with each release. Or it won't even get released. There is still a 1.0.0-SNAPSHOT in the samples release pom.

I will just go for it. For the users this should be irrelevant and if it's improvable for us as maintaners, we can easily change it without affecting users.

PR #775, opened 2022-05-26
drop unneeded config line

hennr:
I don't see what this config line is supposed to do, it tries to assign a field of an interface and does not work as it does not exist after all.


hennr:
FYI @PH1P5 @gesabehrends

Issue #306, opened 2018-11-07
Strategy params column is too short


vargaendre:
Hi,

first, thanks for maintaing this great tool!

I ran into a small issue. I wanted to add ~500 users to a username based strategy. In our system the username is the email address. The comma separated list of the usernames is roughly 10K characters, but the DB column only allows 2000. I changed the column type to text, which allows bigger length, but I think this could be done automatically.
I use the JDBCStateRepository.

Cheers,
Endre

chkal:
Thanks for bringing this up. Back then I decided to go with a VARCHAR column, because I wasn't sure if there is a TEXT data type which works across all major databases.

vargaendre:
Sure. Thanks for the quick response.
I use postgres, btw.

PR #781, opened 2022-06-03
Use text columns on postgres

napstr:
Proposed fix for #306


bennetelli:
@napstr i just reduced the visibility of some methods. Please let me know if you ware fine with it. The rest looks good to me :) Thanks for your PR :)

napstr:
Adjustments look good to me as well 👍

Issue #760, opened 2022-05-12
think about the request-holder-filter for spring boot applications again


bennetelli:
At the moment I am not sure if I would like that users needs to add the request-holder-filter entry in the application.properties. 
Maybe we should discuss this once again and flip the status, so only really edge case users like people who are building an application as a war file and deploying it to an external tomcat instead of using the internal one need to set that entry
and all the others (which I am expecting are far more) who are using Spring Boot as far jar containing everything don't have to set anything like this and the application is just running after they added the starter dependency.

that's the corresponding line to search for in the code: @ConditionalOnProperty(prefix = "togglz.web", name = "request-holder-filter", havingValue = "true")

hennr:
Yep, I like the idea and would vote to have it enabled per Default.
It's quite a bummer when you are trying to use the header activation strategy for example and it simply doesn't work out of the box.

PH1P5:
Hi all,

Is there a need for conditional initialization of the _HttpServletRequestHolderFilter_ within the spring boot starter?
If so I see two options:
- conditional on bean _ActivationStrategyProvider_, since the request holder is only used with the activation strategies
- conditional on web app, since this is a web application feature

wdyt?

bennetelli:
> Yep, I like the idea and would vote to have it enabled per Default. It's quite a bummer when you are trying to use the header activation strategy for example and it simply doesn't work out of the box.

indeed. We should have realized this much earlier.

hennr:
It seems that the default for the request-holder-filter was set to false via PR #537 which was merged to fix the issue #535
This renders the Header activation strategy useless per default which is a bad experience for users and also was not the goal discussed in #535 

I'd suggest to simply turn around the default for the property `togglz.web.request-holder-filter` to true like so:
`@ConditionalOnProperty(prefix = "togglz.web", name = "request-holder-filter", havingValue = "true", matchIfMissing = true)`

This way users would be able to use the Header AS per default and users with WAR deployments would still be able to disable the feature.
Opinions on this @bennetelli  @PH1P5 ?

PH1P5:
I agree having request-holder-filter enabled by default.

PR #793, opened 2022-06-10
Change requet holder filter configuration

bennetelli:
#760


bennetelli:
ich traue dem Braten irgendwie noch nicht so recht. Lasse den PR mal offen und denke nochmal drüber nach ob das so erwartungskonform ist.

spring-boot/starter/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzAutoConfiguration.java
```diff
@@ -255,7 +255,6 @@ public TogglzDialect togglzDialect() {
     @Configuration
     @ConditionalOnClass({OncePerRequestFilter.class})
     @ConditionalOnMissingBean(HttpServletRequestHolderFilter.class)
-    @ConditionalOnProperty(prefix = "togglz.web", name = "request-holder-filter", havingValue = "true")
```
hennr, line 258: This is acutally the place I would add this:

@ConditionalOnProperty(prefix = "togglz.web", name = "request-holder-filter", havingValue = "true", matchIfMissing = true)

This would make the filter enabled per default, so all config changes made by you in this PR in the samples project are fine, but we would probably need to set the config value to false for the war sample application?

benchmarks/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hennr, line 1: should we add this to the .gitignore?

hennr: done

hennr:
> ich traue dem Braten irgendwie noch nicht so recht. Lasse den PR mal offen und denke nochmal drüber nach ob das so erwartungskonform ist.

If you could think about this topic next, that would be call.
We could also have a call about this, if you like.

PR #845, opened 2022-09-28
drop hazelcast integration

bennetelli:


kevinstembridge:
Hi,
I was just about to start using Hazelcast as a state repository. I can't find any discussion indicating why it is being dropped. Is there a recommended alternative for sharing toggle state across instances?

bennetelli:
Hey @kevinstembridge sorry for my late response. 

I started a discussion multiple times and asked if someone needs integration X and Y, but did not get a response. 
I will add it again if you need it. Sorry.. I don't know a better approach then just deleting it and see if someone still needs it :( 

togglz 4.0.0 will contain it again.

kevinstembridge:
I use that approach myself all the time :) 

Thanks for reverting. 

I'm sorry I didn't see any previous discussion. What should I be subscribing to in order to see these threads? I don't really want to see all activity but if you're asking us questions I'd be happy to respond where I can offer something relevant.

Issue #849, opened 2022-09-29
JavaScripting API: encoding issue


BeLi4L:
## Context:

I upgraded from Togglz version 2.8.0 to 3.2.1, but can't edit JavaScript scripts anymore, since every special character gets encoded.

E.g. `'` becomes `&#39;`, `<` becomes `&lt;`, etc. 

## What I did:

![image](https://user-images.githubusercontent.com/8281983/193040895-579f4dd8-51bd-4de6-bffc-877b9062f6f0.png)

## Actual result:

![2022-09-29_15-04](https://user-images.githubusercontent.com/8281983/193040172-43a4576b-4f60-4e4a-8f82-8a9c68b30f86.png)

# Expected result:

![2022-09-29_15-11](https://user-images.githubusercontent.com/8281983/193040472-c74f30ac-77d8-4f93-9b47-e8208b877a37.png)

# Explanation

This may have been introduced by https://github.com/togglz/togglz/pull/449/files#diff-7f7c17da3ef3b2dd9fba8c230694bfcfae96fa0e6d98c376730ed5c73484bc64R61. This prevents me from upgrading at all :sweat_smile: 

I'm not familiar with the Togglz codebase, but what was the XSS bug that MR tried to fix exactly? :thinking:

bennetelli:
Argh.. as far as I remember correctly I just copied the changes from @JoeBeeton. Maybe he can tell us more about it?

BeLi4L:
Here is a PR that should fix this bug (I hope?): https://github.com/togglz/togglz/pull/856

I also rolled back the "Mandatory Java 11" changes if that's OK :yum:

BeLi4L:
@bennetelli Thank you for merging my PR :wink: 

By any chance, do you already have an estimate of when the next release is gonna occur?

BeLi4L:
Wait, actually I think my PR doesn't solve the core issue. It is properly displayed in the console UI, but not stored properly in DB, so the JS is not executed properly.

I should sanitize the input, not escape it entirely. I'll do another PR :wink:

BeLi4L:
Fixed in https://github.com/togglz/togglz/pull/858

PR #856, opened 2022-10-04
Fix JavaScript API encoding issue

BeLi4L:
Fixes https://github.com/togglz/togglz/issues/849


BeLi4L:
> Do you need Java 8 for it? Will need to checkout the PR and test it locally. Maybe I will find some time to do it this evening.

I don't need Java 8 for the fix to work, but I need the fix in the next Java 8-compatible release ^^'

Besides, the Java 8 build fails on Github Actions, due to something being compiled using Java 11, but I have no idea why. Any clue? :thinking:

BeLi4L:
Actually it looks like it's GraalVM 22.2.0, therefore I rolled back to 22.0.0 which looks OK ¯\\\_(ツ)\_/¯

bennetelli (approved):

PR #858, opened 2022-10-06
Fix JavaScript encoding issue

BeLi4L:
Actually fixes https://github.com/togglz/togglz/issues/849

The issue was: we encoded the JS script before saving it, therefore making it corrupted (& became &amp;, etc.).

Now, we store the raw JS sent by the client in DB, and only when DISPLAYING the value, we sanitize it to prevent XSS.

Please tell me if there's another XSS issue somewhere, because apart from the main page, I didn't think of anything else ;)

### DIFF
--- a/console/src/main/resources/org/togglz/console/index.html
+++ b/console/src/main/resources/org/togglz/console/index.html
@@ -60,7 +60,7 @@
                   <ul class="params">
                   ${foreach feature.strategy.parameters param}
                     ${if param.hasValue}
-                      <li>${param.label}: ${param.value}</li>
+                      <li>${param.label}: ${param.value;sanitizeHtml}</li>
                     ${end}
                   ${end}
                   </ul>
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -97,5 +97,10 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
+      <artifactId>owasp-java-html-sanitizer</artifactId>
+      <version>20220608.1</version>
+    </dependency>
   </dependencies>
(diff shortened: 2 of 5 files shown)


BeLi4L:
@bennetelli If you have any questions, please don't hesitate to ask :wink: 

Just to be clear, this adds a way to sanitize the HTML passed to the template engine, using the `${param;sanitizeHtml}` syntax instead of the raw `${param}`. I didn't find any docs on doing this XSS-prevention directly within JMTE, which is really a shame for a template engine, even a "minimal" one :sweat_smile:

console/src/main/resources/org/togglz/console/index.html
```diff
@@ -60,7 +60,7 @@
                   <ul class="params">
                   ${foreach feature.strategy.parameters param}
                     ${if param.hasValue}
-                      <li>${param.label}: ${param.value}</li>
+                      <li>${param.label}: ${param.value;sanitizeHtml}</li>
```
bennetelli, line 63: @BeLi4L does this work? That looks wrong somehow

BeLi4L: Yet that's the official JMTE syntax:
- https://www.tinymediamanager.org/docs/jmte#renderer-information
- https://code.google.com/archive/p/jmte/wikis/BestOfAPI.wiki

It uses the renderer name (here `sanitizeHtml`) defined in `SanitizeHtmlRender.java`

BeLi4L: And yes it works, I tested on my machine, the custom renderer is called, and it prevents XSS on the main page, when displaying the script ;)

bennetelli: Okay thanks :) the syntax looks a bit weird.
Thanks for your PR :)

Issue #870, opened 2022-10-21
Add metadata to Spring Boot Actuator


ptriller:
Hello
We use the spring boot Actuator for administrating our applications, and to automatically generate a GUI similar to the console it would be nice if the Metadata would be available in the actuator endpoint.
I am willing to do the coding myself but I don't want to decide for you how this additional data should be handled.  There are 3 Ways I see:
 - Just add the metadata to the endpoints, the output will not be so much bigger that it might be a problem
 - Add a parameter to the endpoints (e.g ?metadata=true) and only add the data if this parameter is present. The disadvantage here is that while you can just not set the parameter in the web endpoint , there are no default parameters in JMX so the API for jmx would change
 - Just add the metadata to the sub endpoints where only the data for a single featuretoggle is fetched.
 
 for the data structure I propose to add a nested object "metadata":
 
```
{
  "name":"FEATURE_FOUR",
  "enabled":false,
  "strategy":null,
  "params":{}
  "metadata": {
    "label": "Feature Four",
    "groups": [ "group1", "group2" ],
    "enabledByDefault": true,
    "attributes": {}
  }
 ```
  If you tell me what implementation you would be willing to merge I would code it.
  
  
  Regards
  
  Peter

hennr:
Hello @ptriller ,

thanks!
I would be fine with just adding a metadata object.
What do you think @bennetelli ?

When you build this, please also have the test coverage in mind.

PR #871, opened 2022-10-23
Metadata in actuator and tests and fixes

ptriller:
Hello
I implemented integrating the MetaData into the actuator endpoint. (as mentioned [here](https://github.com/togglz/togglz/issues/870))
While doing that I found out that the error handling for the Web Endpoints was not optimal. The Actuator returned a 500 error when trying to set parameters with and illegal format. Since that is a client error and not a server error a 5xx error is inappropriate. For the JMX endpoint throwing an exception is probably ok, but for the web the correct HTTP status code would be better.

I added an `@EndpointWebExtension` to handle validation and errors properly in the HTTP case.
I also added some Integration Tests that test the actual Endpoints. I am not 100 % sure those are necessary, because now basically the same things are testet over 3 channels: JUNIT directly, HTTP and JMX. 

Let me know if I should keep the tests or reduce/remove some of them and if there are any issues with my implementation


bennetelli:
hey @ptriller first of all, thank you very much for your PR :-) 
Basically it's looking good. I would like to bring together the FeatureEnumMetaData interface and the FeatureMetaData class in kotlin with your PR. Does it make sense to you to let your Metadata class implement the FeatureMetaData interface? Just took a brief look at it and I think these classes and interfaces somehow belong together and your metadata class should implement the interface. Does it make sense to you?

ptriller:
> hey @ptriller first of all, thank you very much for your PR :-) Basically it's looking good. I would like to bring together the FeatureEnumMetaData interface and the FeatureMetaData class in kotlin with your PR. Does it make sense to you to let your Metadata class implement the FeatureMetaData interface? Just took a brief look at it and I think these classes and interfaces somehow belong together and your metadata class should implement the interface. Does it make sense to you?

Conceptually the contain the same data , however the `TogglzFeature` and `TogglzFeatureMetaData` are designed to serialize nicely into JSON without any further configuration.
If we add the Interface to the class it would start using `FeatureGroup` and `FeatureState` and both do not easily serialize into clean JSON.
Without adding a dependency to Jackson there is no way to configure how it would be serialized (as far as I know)

As the code is now `TogglzFeature` and `TogglzFeatureMetaData` are purely used as Containers for the data for the acutator endpoint and personally I do not think there is much to gain if we add the Interface. There is no functionality gained by using the `TogglzEndpoint` anywhere else in place of the `FeatureManager` directly. 
However I do not really have an overview over the architecture of Togglz, so if you think it would make things more consistent, I can try to add the Interface.

hennr:
I will try to have a look tomorrow on this!

hennr:
Hi @ptriller , thanks for the PR!

Did I understood it right, that the WebExtension will now handle all http requests and the TogglzEndpoint will handle stuff like JMX?

ptriller:
> Hi @ptriller , thanks for the PR!
> 
> Did I understood it right, that the WebExtension will now handle all http requests and the TogglzEndpoint will handle stuff like JMX?

Correct, since the types need to be a bit different between those two you can make separate custom code for Web/JMX with this WebExtension

hennr:
Works for me so far. I would merge it and add some more comments as wel as fixing the logging setup.

Fine for you, @bennetelli ?

I will be back online next monday.

hennr:
@bennetelli , good to merge if the build was green?

ptriller:
@bennetelli  I would fix it but I am confused. Throughout Master the slf4j version is 1.7.36 which would require a loggack version of 1.2.x however on your master the togglz.logback.version is set to 1.3.4. In other places this is ignored and set to 1.2.11 from somewhere
Should I override the global setting or fix it or wait for a fix ?

hennr:
Hi @ptriller , that's probably my bad.
SLF4J 2.x was used along with logback 1.3.x but this didn't play well with spring-boot 2.x, so we rolled back to slf4j 1 leaving logback at 1.3.x.
With spring-boot 3.x support in place, we need to bump slf4j to 2.x and logack to something like 1.3.x or 1.4.x then.

I rolled back to logback 1.2.11 now, hope this helps with a green build.

hennr:
Finally merged, thanks @ptriller 
Will try to get a new release out tomorrow.

PR #872, opened 2022-10-24
Move enable and disable methods to feature manager

bennetelli:


bennetelli:
oops. the shiro revert is also part if tihs PR

bennetelli:
@frankbregulla1111 would this be fine for you?

PR #888, opened 2022-10-31
Implemented more fine-grained Spring Boot starters containing only required dependencies

cfranzen:
The currently existing togglz-spring-boot-starter module contains a lot of dependencies, like togglz-console and thymeleaf. So even if a project does not use any of these features, the classpath is polluted with these dependencies. This pollution might lead to Spring Boot configuring and starting up some of these dependencies, which is typically not desired. Furthermore, even we using the togglz-spring-boot-starter it is required to additional add further dependencies to a project to activate certain features.

I implemented several more fine-grained Spring Boot starters that bring only dedicated dependencies with them. So only the  tooglz-thymeleaf-spring-boot-starter will have a dependency on thymeleaf. Furthermore, these starters ensure that all required dependencies are add to the classpath, so no more dependencies besides the starters must be configured.

Side note: the currently available togglz-spring-boot-starter has been slightly modified but should be entirely downwards compatible.


spring-boot/autoconfigure/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzAutoConfiguration.java
```diff
@@ -242,7 +243,7 @@ public void addInterceptors(InterceptorRegistry registry) {
     }
 
     @Configuration
-    @ConditionalOnClass(TogglzDialect.class)
+    @ConditionalOnClass(IProcessorDialect.class)
```
bennetelli, line 246: Does this remove the warning during application startup?

cfranzen: Not really. This is just a bug. Having a @ConditionalOnClass with a class that is part of the same module makes no sense at all. I changed it to test for some class that is part of the thymeleaf library.

The warnings actually is fixed by avoiding that the thymeleaf library is on the class path. So using a Spring Boot starter that does NOT contain this library will remove the warning.

bennetelli: I think it was me who implemented it this way :-D interesting. I never thought about this so far. Thanks for the hint. Learned something new :)

hennr:
Hi,

not sure I'd I would like to have so many starters to pick from.
Actuator for example sounds like a perfectly sane default to have, so I'd rather not have a starter for it.
Other proposed starters only duplicate togglz- deps that are basically already there. 

Thymeleaf on the other hand should be left out of the standard spring boot starter, agreed.

So let's start simple and drop the thymeleaf dep from the default starter and make it an own starter.
Then we can see how this fits everybodies needs.
What do you say @cfranzen , @bennetelli ?

cfranzen:
Thanks for your comments @hennr, however I do not agree with your suggestion for several reasons:

- It is not a good practice to have code within a Spring Boot starter project but instead you should have a project containing all autoconfiguration and starters that only combine different dependencies. That is what I tried to achieve with the new autoconfiguration sub-module.
- I did not modify the existing Spring Boot starter by intention to ensure downwards compatibility of my commit. So the classpath of existing users should not changed except that I moved the autoconfiguration code to an additional module (but used the same Java package name)
- Actually I dont care much which Spring Boot starters existing finally except the core starter. In modern (micro-)service oriented architectures you try to avoid exposing services for security reasons. So there is no possibility to access a console or an actuator. For me the sane default should be to expose no togglz endpoint at all and that is what I aimed for with the core starter.
- I do not agree with your comment that the starters only duplicate togglz deps. None of the has only one dependency. But in the end this is just convenience for the togglz user. The togglz user with a Spring Boot app does not want to find out which dependencies are required, to trigger the autoconfiguration. So for me there should be a connection between the available starters and the possibilities to do autoconfiguration. Thats the reasons for my proposed set of starters. They are just in sync with the offered autoconfigurations.

hantsy:
I am using Spring WebFlux stack, so console, web is no use for our project. I also hope adding reactive variant for StateRepository, etc.

spring-boot/autoconfigure/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-core</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
```
cfranzen, line 95: This dependency should not be required since you already have a dependency on spring-boot-starter-web

spring-boot/autoconfigure/src/main/java/org/togglz/spring/boot/actuate/autoconfigure/TogglzAutoConfiguration.java
```diff
@@ -72,7 +74,7 @@
  *
  * @author Marcel Overdijk
  */
-@AutoConfiguration
+@AutoConfiguration // TODO check if this is merged correctly
```
cfranzen, line 77: Should this TODO be removed?

cfranzen (commented):
Some minor comments

spring-boot/autoconfigure/pom.xml
cfranzen: Sorry, did not see that spring-boot-starter-web is in test scope

PR #896, opened 2022-11-05
Remove deprecated methods

bennetelli:
Merge this if we want to release togglz 4.0.0


hennr:
I'd like to take a look as well before dropping.

core/src/main/java/org/togglz/core/repository/FeatureState.java
```diff
@@ -119,50 +100,6 @@ public FeatureState disable() {
...
-    /**
-     * The list of users associated with the feature state.
-     *
-     * @return The user list, never <code>null</code>
-     *
-     * @deprecated This method will be removed soon. Use {@link #getParameter(String)} instead to read the corresponding
-     *             strategy parameter.
-     */
-    @Deprecated
-    public List<String> getUsers() {
```
hennr, line 131: All three users methods seem quite handy in comparison to the get/setParameters method that only accepts a string.
Any good reason to drop them really, @bennetelli ?

PR #909, opened 2022-11-28
Reduce allocations from DefaultMapSerializer

dreis2211:
Hi,

I've been profiling some test suites lately and noticed that Togglz does a small, but noticeable, amount of allocations when reloading feature states before & after tests:

E.g. see the following excerpts. Overall, all of the different frames account for ~1% of allocations.
image
image

The unfortunate bit here is that the work in `DefaultMapSerializer` could be largely skipped if the Map is empty (which it is in most default cases). So this PR checks for emptiness on (de-)serialization to save some work & allocations.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/core/src/main/java/org/togglz/core/repository/util/DefaultMapSerializer.java
+++ b/core/src/main/java/org/togglz/core/repository/util/DefaultMapSerializer.java
@@ -14,7 +14,7 @@ import java.util.Properties;
 
 /**
  * This converter is able to convert string maps to simple strings and vice versa.
- * 
+ *
  * @author Christian Kaltepoth
  */
 public class DefaultMapSerializer implements MapSerializer {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/togglz/core/repository/util/DefaultMapSerializer.java
```diff
@@ -59,16 +59,15 @@ public static DefaultMapSerializer multiline(String lineSeparator) {
         return new DefaultMapSerializer(true, lineSeparator);
     }
 
-    /*
-     * (non-Javadoc)
-     * 
-     * @see org.togglz.core.util.MapSerializer#convertToString(java.util.Map)
```
dreis2211, line 65: I felt free to remove this as it doesn't seem to exist anymore. Same below

bennetelli (approved):
lgtm

PR #914, opened 2022-12-16
Reduce overhead of TogglzTestExecutionListener

dreis2211:
Hi,

I've noticed that our test-suite spends around ~3-4% of CPU cycles just inside `TogglzTestExecutionListener`.

image

The unfortunate bit here is that we don't even customize the feature flags in the majority of cases and deal with the default state. It seems that we could skip setting the feature flag state in those cases - which resorts to the JDBC implementation for us - and avoid the corresponding overhead. The additional benefit in our case is that the `CachingStateRepository` will have more cache hits and avoid the reading overhead from the database as well.

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/core/src/main/java/org/togglz/core/repository/FeatureState.java
+++ b/core/src/main/java/org/togglz/core/repository/FeatureState.java
@@ -8,6 +8,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 
 import org.togglz.core.Feature;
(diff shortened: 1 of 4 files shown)


dreis2211:
Fixed the failing `InMemoryStateRepositoryTest`

bennetelli:
thanks for your PR @dreis2211 :) I am back from vacation and finished all my talks and workshops. So I will have time to work on your PRs this week.

bennetelli (approved):
makes sense to me as well. lgtm

thanks for your PR :)

dreis2211:
Any chance this could be backported or released before 4.0.0 with all the breaking changes in it?

Issue #14, opened 2013-02-11
 Column length too big in MySQL


ghost:
hi, when i try to use the datasource repository option (with mysql 5.2.42) i get the following error:
"Column length too big for column 'FEATURE_USERS' (max = 255); use BLOB or TEXT instead"

than i have 2 options:
1.override the JDBCStateRepository file and change the either to "FEATURE_USERS TEXT(2000)" or to "FEATURE_USERS CHAR(255)"
2.create the table manuali myself

any help will be appreciated

PR #917, opened 2023-01-11
Jakarta, Spring Boot 3 and Spring 6

MediaMarco:
This is a pull request for a possible 4.x (?) version which includes the migration to the `jakarta` namespace. Spring Boot 3 and Spring 6 are working, Java 17 or higher is required.

Currently excluded modules:

Cassandra (Netflix Astyanax is retired, Cassandra Thrift is gone in latest Cassandra version, this module needs some work)
DynamoDB (Uses AWS SDK for Java v1 which only supports Java up to 16, has to be migrated to AWS SDK for Java v2)


pom.xml
```diff
@@ -95,7 +106,7 @@
                 </configuration>
             </plugin>
 
-            <plugin>
+            <!--<plugin>
```
OleksandrShkurat, line 109: I've found that jacoco can work here in caes of configuring it with the following:
`                <configuration>
                    <includes>
                        <include>org/togglz/**</include>
                    </includes>
                </configuration>
`

appengine/pom.xml
```diff
@@ -35,12 +35,11 @@
       <version>${appengine.target.version}</version>
     </dependency>
 
-    <dependency>
-      <groupId>javax.inject</groupId>
-      <artifactId>javax.inject</artifactId>
-      <version>1</version>
-      <scope>provided</scope>
-    </dependency>
+      <dependency>
```
OleksandrShkurat, line 38: 1) In my case it was enough to replace this dependency with `jakarta.inject:jakarta.inject-api:2.0.1` keepnig the same scope (provided)
2) It would be great to keep existing alignment for the dependency block (2 spaces for indention instead of 4)

cdi/pom.xml
```diff
@@ -20,17 +20,22 @@
...
       <scope>provided</scope>
     </dependency>
 
-    <dependency>
-      <groupId>javax.enterprise</groupId>
-      <artifactId>cdi-api</artifactId>
-      <version>2.0.SP1</version>
-      <scope>provided</scope>
-    </dependency>
+      <dependency>
```
OleksandrShkurat, line 28: 1) jakarta.enterprise.cdi-api:4.0.1 should be used instead
2) it gives transitive dependency for jakarta.annotation-api as well making it not necessary to define it explicitly

cdi/src/main/java/org/togglz/cdi/spi/CDIBeanFinder.java
```diff
@@ -67,6 +68,21 @@ private BeanManager getBeanManager(ServletContext servletContext) {
...
         }
 
+        // try Weld 5.1.x servlet context attribute
+        if (beanManager == null && servletContext != null) {
+            beanManager = (BeanManager) servletContext.getAttribute(SERVLET_CONTEXT_ATTR_WELD_5_1);
+        }
+
+        // try jakarta servlet context attribute
+        if (beanManager == null && servletContext != null) {
+            beanManager = (BeanManager) servletContext.getAttribute("jakarta.enterprise.inject.spi.BeanManager");
```
OleksandrShkurat, line 78: could you please extract this string to a constant as well?

cdi/src/main/java/org/togglz/cdi/spi/CDIBeanFinder.java
```diff
@@ -67,6 +68,21 @@ private BeanManager getBeanManager(ServletContext servletContext) {
...
+        }
+
+        // try jakarta servlet context attribute
+        if (beanManager == null && servletContext != null) {
+            beanManager = (BeanManager) servletContext.getAttribute("org.jboss.weld.manager.BeanManagerImpl");
```
OleksandrShkurat, line 83: could you please extract this string to a constant as well?

cdi/src/main/java/org/togglz/cdi/spi/CDIBeanFinder.java
```diff
@@ -21,6 +21,7 @@ public class CDIBeanFinder implements BeanFinder {
     public final static String BEAN_MANAGER_JNDI_TOMCAT = "java:comp/env/BeanManager";
 
     public final static String SERVLET_CONTEXT_ATTR_WELD_1_1 = "org.jboss.weld.environment.servlet.javax.enterprise.inject.spi.BeanManager";
+    public final static String SERVLET_CONTEXT_ATTR_WELD_5_1 = "org.jboss.weld.environment.servlet.jakarta.enterprise.inject.spi.BeanManager";
```
OleksandrShkurat, line 24: I'd suggest reordering access modifiers here and for existing constants as well to comply with the Java Language specification (`public static final`)

cloud-datastore/pom.xml
```diff
@@ -36,6 +36,12 @@
             <artifactId>togglz-core</artifactId>
         </dependency>
 
+        <dependency>
```
OleksandrShkurat, line 39: please keep the original scope (`provided`)

cdi/src/main/java/org/togglz/cdi/spi/CDIBeanFinder.java
MediaMarco: I removed the block, this did not work anyway.

cdi/src/main/java/org/togglz/cdi/spi/CDIBeanFinder.java
MediaMarco: I removed the block, this did not work anyway.

spring-boot/pom.xml
```diff
@@ -27,7 +27,7 @@
             <dependency>
                 <groupId>org.springframework</groupId>
                 <artifactId>spring-core</artifactId>
-                <version>5.3.23</version>
+                <version>${spring.version}</version>
```
OleksandrShkurat, line 30: I'd suggest removing the version at all. Let spring-boot define it itself

shiro/src/test/java/org/togglz/shiro/test/ShiroLoginServlet.java
```diff
@@ -19,7 +19,7 @@ public class ShiroLoginServlet extends HttpServlet {
     private static final long serialVersionUID = 1L;
 
     @Override
-    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
OleksandrShkurat, line 22: `doGet` method still exposes `ServletException`. I beleive it shouldn't be removed.
I've rechecked all three instances available in the project:
- jakarta.servlet:jakarta.servlet-api:6.0.0
- jakarta.platform:jakarta.jakartaee-web-api:10.0.0
- org.apache.tomcat.embed:tomcat-embed-core:10.1.4

spring-boot/starter/pom.xml
```diff
@@ -1,156 +1,170 @@
...
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-configuration-processor</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.thymeleaf</groupId>
+            <artifactId>thymeleaf</artifactId>
+            <version>3.1.1.RELEASE</version>
```
OleksandrShkurat, line 56: I'd suggest removing both explicitly defined versions of thymeleaf. Spring-boot defines the same version for them

spring-core/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.togglz</groupId>
     <artifactId>togglz-project</artifactId>
-    <version>3.2.1-SNAPSHOT</version>
+    <version>4.0.0-SNAPSHOT</version>
```
OleksandrShkurat, line 8: `spring-core` module still contains `spring,factories` file not converted to `org.springframework.boot.autoconfigure.AutoConfiguration.imports`

test-harness/src/main/resources/arquillian.xml
```diff
@@ -2,15 +2,16 @@
...
   <engine>
     <property name="deploymentExportPath">/tmp</property>
   </engine>
   -->
 
   <container qualifier="wildfly-managed" default="true">
     <configuration>
-      <property name="jbossHome">../target/container/wildfly-22.0.1.Final/</property>
+      <property name="jbossHome">../target/container/wildfly-27.0.1.Final/</property>
+      <!--<property name="jbossArguments">-Dsome.property=bla</property>-->
```
OleksandrShkurat, line 14: looks like the line #14 was commited occasionally

spring-core/pom.xml
MediaMarco: According to https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/TestExecutionListener.html the spring.factories file is still the way to go for TestExecutionListener registration in Spring 6.

samples/spring-boot-security/src/main/java/sample/Application.java
```diff
@@ -2,57 +2,45 @@
...
-    protected static class ApplicationSecurity extends WebSecurityConfigurerAdapter {
+    protected static class ApplicationSecurity {
 
-        @Override
-        protected void configure(HttpSecurity http) throws Exception {
+        @Bean
+        protected SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
             // @formatter:off
             http
                     .authorizeRequests()
```
OleksandrShkurat, line 23: I'd suggest replacing the deprecated method `authorizeRequests` with suggested replacement for it as `authorizeHttpRequests`

samples/spring-boot-security/src/main/java/sample/Application.java
```diff
@@ -2,57 +2,45 @@
...
-                        .anyRequest()
-                        .authenticated()
-                        .and()
+                    .anyRequest()
```
OleksandrShkurat, line 24: Could you please revert formatting in this block to the keep original one as it was requested by its author @bennetelli using @formatter:off hint ?

samples/spring-boot-security/src/main/java/sample/Application.java
```diff
@@ -2,57 +2,45 @@
...
-                        .password("pwd")
-                        .roles("USER")
-                        .and()
-                    .withUser("user4")
-                        .password("pwd")
-                        .roles("USER");
-            // @formatter:on
+        @Bean
+        public InMemoryUserDetailsManager userDetailsService() {
+            UserDetails admin = User.withDefaultPasswordEncoder().username("admin").password("pwd").roles("ADMIN", "USER").build();
```
OleksandrShkurat, line 38: I'd suggest keeping this method formatting/semantics closer to the original one. For example:

```
            // @formatter:off  
            UserDetails admin = User.withDefaultPasswordEncoder()  
                                    .username("admin")  
                                    .password("pwd")  
                                    .roles("ADMIN", "USER")  
                                    .build();  
            UserDetails user1 = User.withDefaultPasswordEncoder()  
                                    .username("user1")  
                                    .password("pwd")  
                                    .roles("USER")  
                                    .build();  
            UserDetails user2 = User.withDefaultPasswordEncoder()  
                                    .username("user2")  
                                    .password("pwd")  
                                    .roles("USER")  
                                    .build();  
            UserDetails user3 = User.withDefaultPasswordEncoder()  
                                    .username("user3")  
                                    .password("pwd")  
                                    .roles("USER")  
                                    .build();  
            UserDetails user4 = User.withDefaultPasswordEncoder()  
                                    .username("user4")  
                                    .password("pwd")  
                                    .roles("USER")  
                                    .build();  
            // @formatter:on  
```

spring-core/pom.xml
MediaMarco: It's also still mentioned here: https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#testcontext-tel-config-automatic-discovery

There's also a statement on https://stackoverflow.com/questions/73317916/spring-boot-2-7-migrating-spring-factories-into-imports-file which states that spring.factories is deprecated for auto-configuration classes and test slices, but not for registering TestExecutionListener instances for the tests.

samples/spring-boot-security/src/main/java/sample/Application.java
OleksandrShkurat, line 38: The method `withDefaultPasswordEncoder` is marked as deprecated where the explanation says that it recommended just for demo purposes but not for production. We use it exactly for demo purposes so my suggestion is to add annotation `@SupressWarnings("deprecation")` on top of our method `userDetailsService`

samples/pom.xml
```diff
@@ -29,7 +29,7 @@
         <module>spring-boot-no-feature-provider-or-feature-enums</module>
         <module>spring-boot-security</module>
         <module>spring-boot-thymeleaf</module>
-        <module>spring-boot-war</module>
+        <!--<module>spring-boot-war</module> TODO: Fix war sample -->
```
OleksandrShkurat, line 32: I've fixed spring-boot-war by defining
```
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-war-plugin</artifactId>
                    <version>3.3.2</version>
                </plugin>
```
in `pluginManagement` section of `\samples\pom.xml`

After that it became useless to define explicitly the version of this plugin in `\samples\spring-boot-hello-world\pom.xml`

MediaMarco: Great, thank you!

OleksandrShkurat:
I've found that the project contains an orphaned module `microprofile-config` which is not reference from amywhere. I've discovered that its usage was introduced in 2017 and then removed in 2020. So that, looks like the whole folder with rests of this orphaned module could be easily just removed. What do you think, is it worth enough to do this in a scope of this PR?

samples/spring-boot-hello-world/pom.xml
```diff
@@ -66,17 +66,8 @@
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
             </plugin>
-
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-war-plugin</artifactId>
-                <version>3.3.2</version>
```
OleksandrShkurat, line 73: I suggested to remove just version number keeping everithing else without changes. Does it work actually?

samples/pom.xml
OleksandrShkurat: I suggested to remove just version number from `\samples\spring-boot-hello-world\pom.xml` keeping the plugin itself with its settings without changes. Does it work actually when the whole plugin reference removed?

pom.xml
MediaMarco: Jacoco works again, just commented in in again. Was broken during migration, but forgot to re-enable it.

OleksandrShkurat:
Another question to discuss. While trying to upgrade the project I also upgraded many used dependencies and plugins to more recent versions. The build with all of them on my local machine is green. Should I comment appropriate POMs?

amazon-s3/pom.xml
```diff
@@ -28,7 +28,7 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>s3</artifactId>
-            <version>2.18.35</version>
+            <version>2.19.8</version>
```
OleksandrShkurat, line 31: `2.19.12` is already available. Is there any concern for using the most recent version?

cloud-datastore/pom.xml
```diff
@@ -14,7 +14,7 @@
   <description>Togglz - Google Cloud Datastore Integration</description>
 
   <properties>
-    <gcloud.version>2.11.5</gcloud.version>
+    <gcloud.version>2.13.0</gcloud.version>
```
OleksandrShkurat, line 17: Can we use `2.13.1` instead?

core/pom.xml
```diff
@@ -34,12 +34,12 @@
     <dependency>
       <groupId>org.graalvm.js</groupId>
       <artifactId>js</artifactId>
-      <version>22.0.0</version>
+      <version>22.2.0</version>
```
OleksandrShkurat, line 37: 1) `22.3.0` is already available
2) what about introducing a property (i.e. `graalvm-js.version`) and reuse it twice?

pom.xml
```diff
@@ -493,7 +493,7 @@
                 <!-- Used by proxy generation -->
                 <groupId>net.bytebuddy</groupId>
                 <artifactId>byte-buddy</artifactId>
-                <version>1.12.18</version>
+                <version>1.12.20</version>
```
OleksandrShkurat, line 496: `1.12.21` is available

redis/pom.xml
```diff
@@ -20,12 +20,12 @@
...
+			<version>4.3.1</version>
 		</dependency>
 		<dependency>
 			<groupId>io.lettuce</groupId>
 			<artifactId>lettuce-core</artifactId>
-			<version>6.2.0.RELEASE</version>
+			<version>6.2.2.RELEASE</version>
 		</dependency>
 		<dependency>
 			<groupId>org.testcontainers</groupId>
```
OleksandrShkurat, line 31: Could you also replace `1.17.6` here with `${testcontainers.version}`?

spock/pom.xml
```diff
@@ -19,6 +19,7 @@
         <dependency>
             <groupId>org.spockframework</groupId>
             <artifactId>spock-core</artifactId>
+            <version>2.3-groovy-3.0</version>
```
OleksandrShkurat, line 22: Could you please update this version in the root pom.xml by changing the property `spockframework.version`? It's not neccessary to define exact version here

test-harness/pom.xml
```diff
@@ -19,7 +19,7 @@
         <dependency>
           <groupId>net.sourceforge.htmlunit</groupId>
           <artifactId>htmlunit</artifactId>
-          <version>2.67.0</version>
+          <version>2.67.0</version><!--2.68.0 and 2.69.0 lead to compilation errors TODO: Investigate-->
```
OleksandrShkurat, line 22: `2.69.0` works well for me. Could you please share which exact compilation error do you have? I guess I have solution for it.

dynamodb/pom.xml
```diff
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-dynamodb</artifactId>
-            <version>1.12.321</version>
+            <version>1.12.376</version>
```
OleksandrShkurat, line 23: I've updated it to `1.12.380` locally

OleksandrShkurat:
We have four places where `jackson-databind` is used. Please update them from `2.13.4.2` to `2.14.1`
- amazon-s3
- cloud-storage
- dynamodb
- slack

BTW, oficially even 2.14.1 is not completely ready to JDK17. It's expected that upcoming version 2.15.x will support it fully

servlet/pom.xml
```diff
@@ -1,70 +1,71 @@
 <?xml version="1.0"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
OleksandrShkurat, line 2: Could you please reformat again this XML after those changes performed in `.editorconfig` expecting that indentation should be sized by two spaces instead of default four?

test-harness/pom.xml
```diff
@@ -44,16 +44,28 @@
         <dependency>
             <groupId>org.jboss.arquillian.junit</groupId>
             <artifactId>arquillian-junit-container</artifactId>
-            <version>1.6.0.Final</version>
+            <version>1.7.0.Alpha13</version>
         </dependency>
 
+<!--        <dependency>-->
```
OleksandrShkurat, line 50: these commented dependencies aren't necessary to keep

bennetelli:
wow guys! aweseome what's going on here :) thank you so much for your help and your time you invested in moving this forward :)

OleksandrShkurat:
@bennetelli thank you. What do you think, is the folder `microprofile-config` still needed? As I can see, it's an orphaned part of some old removed logic. Am I right?

MediaMarco:
CDI module works again, the failing test was fixed.

.github/workflows/maven.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jdk: [8, 11]
+        jdk: [17]
```
bennetelli, line 18: do we actually want to set Java 17 as minimum?

cdi/src/test/java/org/togglz/cdi/container/ManagedFeatureManagerProvider.java
```diff
@@ -1,17 +1,17 @@
...
-import javax.enterprise.inject.Produces;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Produces;
 import org.togglz.cdi.Features;
 import org.togglz.core.manager.FeatureManager;
 import org.togglz.core.manager.FeatureManagerBuilder;
 import org.togglz.core.repository.mem.InMemoryStateRepository;
 import org.togglz.core.user.NoOpUserProvider;
 
+@ApplicationScoped
```
bennetelli, line 11: why do we need to put it on the class instead of just keeping it on the produce() method?

spring-boot/starter/src/test/java/org/togglz/spring/boot/actuate/TogglzSpringBootEndpointTest.java
```diff
@@ -63,6 +71,7 @@ public static class TestConfig {
 
 
     @Test
+    //@WithMockUser(username = "username", roles={"ADMIN"})
```
bennetelli, line 74: is this still needed?

bennetelli:
> @bennetelli thank you. What do you think, is the folder `microprofile-config` still needed? As I can see, it's an orphaned part of some old removed logic. Am I right?

I would say 90% of the togglz users is using togglz with Spring Boot, but I could also be wrong.
If we could get it up and running again without much effort, then this would be great. If not, we could also consider to remove it as it takes a lot of time to maintain all the extensions.

.github/workflows/maven.yml
OleksandrShkurat: Good question. Spring boot 3 requires Java 17. However, some other modules could work with older versions

OleksandrShkurat:
> > @bennetelli thank you. What do you think, is the folder `microprofile-config` still needed? As I can see, it's an orphaned part of some old removed logic. Am I right?
> 
> I would say 90% of the togglz users is using togglz with Spring Boot, but I could also be wrong. If we could get it up and running again without much effort, then this would be great. If not, we could also consider to remove it as it takes a lot of time to maintain all the extensions.

Please correct me if I'm wrong but currently, there is no any reference to the folder `microprofile-config` from any place in the project. It was referenced between 2017 and 2020 like this:
```
    <profile>
      <id>java8-features</id>
      <activation>
        <jdk>[1.8,)</jdk>
      </activation>
      <modules>
        <module>microprofile-config</module>
      </modules>
    </profile>
```

and this commit 48e83bb1453d198a6fd842d15a23da4e05bbca4e removed this profile making the folder unused since that time

cdi/src/test/java/org/togglz/cdi/container/ManagedFeatureManagerProvider.java
OleksandrShkurat: One of Marco's commits says:
 Fix CDI module, the @ApplicationScoped annotation has to be at class level to be discovered by new default bean-discovery-mode annotated in CDI 4.0
So, looks like yes, it should be exactly on the class level.

OleksandrShkurat:
Additionally, can we discuss somehow, what is the defintion of done for this activity?
Can we collect all the acceptance criteria for the version 4.0 to be published?

MediaMarco:
Shiro is back!

MediaMarco:
Proposal: Create a 4.x-branch and merge this pull request into it. The open topics that don't really touch the Jakarta / Spring / JDK 17 migration can then be solved as individual pull requests. Wouldn't this be better?

shiro/src/test/java/org/togglz/shiro/test/ShiroLoginServlet.java
bennetelli: not sure if I get it. I would say it's not needed as a checked exception here anymore. If it would, then the compiler should complain about it when it's missing. Maybe they changed it from checked to unchecked exception?

I would not add the ServletException here again. I think there is no benefit of doing it.

.github/workflows/maven.yml
bennetelli: it's actually hard to say if and how many people are using togglz with legacy Java. I have actually no clue about the different use cases of the togglz users.

@MediaMarco uses togglz with Spring Boot. Many others as well. But not sure about the distribution of number of Spring Boot users and number of all togglz users.. :/ 

I just checked it. Quarkus is also on Java 17. So I think we should just got for it. 
Even if I would open a poll or something like that, I would not get enough responses to know if it would be a show stopper to set java 17 as minimum. Let's just do it and see if someone actually needs it to be on a older Java version or not.

cdi/src/test/java/org/togglz/cdi/container/ManagedFeatureManagerProvider.java
bennetelli: Okay. Interesting. Then this is fine.

bennetelli:
> > > @bennetelli thank you. What do you think, is the folder `microprofile-config` still needed? As I can see, it's an orphaned part of some old removed logic. Am I right?
> > 
> > 
> > I would say 90% of the togglz users is using togglz with Spring Boot, but I could also be wrong. If we could get it up and running again without much effort, then this would be great. If not, we could also consider to remove it as it takes a lot of time to maintain all the extensions.
> 
> Please correct me if I'm wrong but currently, there is no any reference to the folder `microprofile-config` from any place in the project. It was referenced between 2017 and 2020 like this:
> 
> ```
>     <profile>
>       <id>java8-features</id>
>       <activation>
>         <jdk>[1.8,)</jdk>
>       </activation>
>       <modules>
>         <module>microprofile-config</module>
>       </modules>
>     </profile>
> ```
> 
> and this commit [48e83bb](https://github.com/togglz/togglz/commit/48e83bb1453d198a6fd842d15a23da4e05bbca4e) removed this profile making the folder unused since that time

That's true. I started a poll some time ago and asked if someone is still using integration A,B and so on. After getting no responses I just disabled some of them to see if someone opens issues.
Looks like nobody complained about the missing microprofile integration.

bennetelli:
> Additionally, can we discuss somehow, what is the defintion of done for this activity? Can we collect all the acceptance criteria for the version 4.0 to be published? We're talking about the major update and it means that some significant changes are expected. Except following:
> 
>     * moving from javax to jakarta
> 
>     * upgrading to JDK17
> 
>     * switch to Spring-boot 3
> 
>     * Spring Framework 6
> 
> 
> we could consider some other changes. For example:
> 
>     * decomissioning of some part of code that was marked as deprecated earlier
> 
>     * complete migration of internal unittests from Junit4 to Junit5 (dependency on Junit4 could be removed, currently is in-progress on my side)
> 
>     * review internal relations among project modules (simplify the dependency tree of the project, I've done it locally)
> 
>     * extend `.editorconfig` with more rules for the code style unifying including import ordering

exactly.

I would also add PRs like this one: #896

---

Be careful when migrating to JUnit5. As far as I remember correctly, the Spock frameworks still depends on JUnit 4. We should keep this in mind

kevinstembridge:
I'm upgrading a side project to Spring Boot 3 and I'm currently blocked by Togglz being incompatible. 
Let me know if I can help with anything here. I'd be happy to test out a 4.0 beta once it is created.

bennetelli:
> I'm upgrading a side project to Spring Boot 3 and I'm currently blocked by Togglz being incompatible. Let me know if I can help with anything here. I'd be happy to try out a 4.0 beta once it is created.

* dynamodb needs to be updated
* cassandra as well

As far as I know

PR #978, opened 2023-03-21
Migrate DynamoDB module to use AWS SDK for Java 2.x

Harmelodic:


dynamodb/src/test/java/org/togglz/dynamodb/DynamoDBStateRepositoryIT.java
```diff
@@ -23,33 +26,33 @@ public class DynamoDBStateRepositoryIT {
...
-    private AmazonDynamoDBClient setupAmazonDbClient() {
-        AmazonDynamoDBClient client = new AmazonDynamoDBClient(
-                new BasicAWSCredentials("", "not_really_used")
-        );
-        client.withEndpoint(String.format("http://localhost:%s", PORT));
-        client.setSignerRegionOverride("us-east-1");
-        return client;
+    private DynamoDbClient setupAmazonDbClient() {
+        return DynamoDbClient.builder()
+                .credentialsProvider(DefaultCredentialsProvider.builder().build())
```
Harmelodic, line 43: I'm pretty confident about everything I've changed, except for this bit.
Tests pass, so... cool?

Harmelodic: You're welcome! Happy to help out.

I realised that when I said tests pass, it was because I was doing `mvn clean tests` on the module and it worked, but this is an IT so runs in `verify`.

GitHub Action tests ended up failing, so I investigated and found that I can define a custom `AwsCredentialsProvider` in a lambda to return predefined `AwsBasicCredentials`.
So, I've setup the credentials in the same way that was before, just now in a lamdbda, since that's how AWS SDK for Java 2.x seems to want to have it handled.

Harmelodic:
I've updated the code now to actual have the tests pass (I was using `mvn clean test` before)

`mvn clean verify` works locally now.

bennetelli:
Thank you very much for your  PR :-)

PR #1033, opened 2023-05-26
togglz-servlet web-fragment namespace value is not compliant with 

franzmathauser:
The value of the namespace of META-INF/web-fragment.xml contains a trailing "/" value. 

This is not compliant with
https://jakarta.ee/xml/ns/jakartaee/web-fragment_6_0.xsd

and will result in an error during application bootstrap if the application server (in my case payara 6) does xml validation. 

To resolve this issue, i changed the value to "https://jakarta.ee/xml/ns/jakartaee" without the trailing "/"


bennetelli:
@franzmathauser thanks for your PR. It's merged already and I am releasing togglz 4.0.1 already. Should be in maven central within the next minutes / hours. Somehow maven central is a bit slow today.

PR #1112, opened 2023-08-24
Remove parameters with empty values.

Jazornowsky:
As far I can see, there is currently no possibility to remove parameter, when it was previously assigned. This change fixes that.


bennetelli:
Hey @Jazornowsky thanks for your PR :) 
Somehow I don't get it. Can't you just call FeatureState.setParameter("your-feature-name", null); to remove the parameter you don't want anymore?

Jazornowsky:
Hey. If im using Spring actuator endpoint and POSTing I can't. If I send null "parameters" or empty value assigned to parameter, no change in parameters are made. TogglzEndpointWebExtension#setFeatureState.

PR #1157, opened 2024-02-05
have a reliable order of features in kotlin-module to fix switching console output

ewoelfel:


bennetelli (approved):
much thanks for your PR, @ErikWoelfel2323 :)

PR #1188, opened 2024-04-22
dependabot pull request limit 20

sullis:


hennr:
Hi @sullis , I think the main problem here is that we have PRs that are a couple of years old.
I don't think that we will increase the throughput of updates if we increase the PR limit for dependabot.

PR #1198, opened 2024-06-28
#1197: advanced test support

### LINKED ISSUE #1197: Improve test support
Quote from your website:

> One of the mission statements of Togglz is that it should be easy to integrate into applications. 

In general I love togglz for its simplicity so thank you for making it.
However, sorry to be frank but I would clearly state that for testing this "mission statement" definitely does not apply
Search the web for "IllegalStateException: Could not find the FeatureManager."
I do not want to be forced to add togglz test dependency to all my modules and pollute all my tests with stuff like `@AllDisabled(MyFeature.class)`.

IMHO you could solve this way smarter for the 95% case.
This example is from your website:
```java
public enum MyFeatures implements Feature {

    @EnabledByDefault
    @Label("First Feature")
    FEATURE_ONE,
    
    @Label("Second Feature")
    FEATURE_TWO;
    
    public boolean isActive() {
        return FeatureContext.getFeatureManager().isActive(this);
    } 
}
```

So why can't togglz work in a way that without any configuration this is just giving me the defaults so that `FEATURE_ONE` is active (`@EnabledByDefault`) and `FEATURE_TWO` is not (in JUnit tests)?
In case I want something else, I am happy to add custom JUnit dependencies and magic annotations.
I just introduced Togglz to a project with million lines of code and 600 maven modules.
Now with more components using feature-toggles all the existing tests fall apart.

I am fully aware that my "feature request" would need some magic fallback for testing in the productive togglz code.
Feel free to require some environment variable that developers could somehow set crosscutting via surefire, spring or whatever frameworks involved if you think there is no other way.
Also feel free to spam the logs with aggressive warnings if togglz is not configured and decides to fall back to this test mode since it somehow assumes it is currently running in a JUnit setup.
All is doable and has already been done by other Java libs.
IMHO this would increase your UX for developers significantly. Thanks for taking care.

hohwille:
Fixes #1197 
IMHO a clean and simple approach.
No documentation included.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -187,6 +187,7 @@
         <module>redis</module>
         <module>kotlin</module>
         <module>samples</module>
+        <module>test</module>
     </modules>
 
     <developers>
--- a/test/pom.xml
+++ b/test/pom.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.togglz</groupId>
+    <artifactId>togglz-project</artifactId>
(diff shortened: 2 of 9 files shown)


hohwille:
Oh, no github-actions for CI of PRs?

hohwille:
Project still alive? Any feedback?

